PR #1, opened 2016-08-07
Upgrade level 1 to level 2

YijinL:
Remaining:
- [ ] make storage work
- [ ] Unit tests
- [ ] System test (i/o redirection) update
- [ ] Update documentation

### DIFF
--- a/src/seedu/addressbook/Main.java
+++ b/src/seedu/addressbook/Main.java
@@ -1,13 +1,15 @@
 package seedu.addressbook;
-/* ==============NOTE TO STUDENTS======================================
- * This class is written in a procedural fashion (i.e. not Object-Oriented)
- * Yes, it is possible to write non-OO code using an OO language.
- * ====================================================================
- */
+
+import static seedu.addressbook.TextUi.*;
+import static seedu.addressbook.StorageFile.InvalidStorageFilePathException;
+import seedu.addressbook.model.AddressBook;
+import seedu.addressbook.Parser.Command;
+import seedu.addressbook.model.InvalidDataException;
+import seedu.addressbook.model.Tag;
+import seedu.addressbook.model.UniqueTagList;
+import seedu.addressbook.model.person.*;
 
 import java.io.*;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.util.*;
 
 /* ==============NOTE TO STUDENTS======================================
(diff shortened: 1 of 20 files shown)


YijinL:
@damithc: fully featured except for storage. 
Will follow the bottom-up testing principle as stated on slack.
Which do we want to use, xml or json?

damithc:
json vs xml, whichever that gives the simpler code.

YijinL:
Followed the architecture diagram at the google docs as close as possible,
except for some renames, and `Person` contains `UniqueTagList`

src/seedu/addressbook/Utils.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public static void assertNotNull(Object... items) {
+        int argIndex = 0;
+        for (Object item : items) {
+            if (Objects.isNull(item)) {
+                throw new AssertionError("Argument at index " + argIndex + " is null");
+            }
+            argIndex++;
+        }
+    }
```
damithc: No need for the code to be defensive. i.e. don't verify assumptions at all. If we are going to do that, might as well use `assert` rather than write our own assert.

src/seedu/addressbook/model/person/UniquePersonList.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * A list of persons that enforces no nulls and uniqueness between its elements.
+ *
+ * Supports minimal set of list operations needed for the app's features.
+ *
+ * @see Person#equals(Object)
+ * @see Utils#elementsAreUnique(Collection)
+ */
+public class UniquePersonList implements Iterable<Person> {
```
damithc: I wonder if we can use `Set` instead of `UniqueList`

src/seedu/addressbook/TextUi.java
```diff
@@ -0,0 +1,327 @@
...
+    public static final String MESSAGE_PERSON_NOT_IN_ADDRESSBOOK = "Person could not be found in address book";
+    public static final String MESSAGE_PERSONS_FOUND_OVERVIEW = "%1$d persons found!";
+    public static final String MESSAGE_WELCOME = "Welcome to your Address Book!";
+    public static final String MESSAGE_USING_STORAGE_FILE = "Using storage file : %1$s";
+    public static final String MESSAGE_VIEW_PERSON_DETAILS = "Viewing person: %1$s";
+
+    public static final String COMMAND_ADD_DESC = "Adds a person to the address book.";
+    public static final String COMMAND_ADD_PARAMETERS = "NAME [p]p/PHONE [p]e/EMAIL [p]a/ADDRESS  [t/TAG]...";
+    public static final String COMMAND_ADD_EXAMPLE = Command.ADD.commandWord + 
+            " John Doe p/98765432 e/johnd@gmail.com a/311, Clementi Ave 2, #02-25 t/friends t/owes_money";
```
damithc: Should we introduce a `Command` class? No need to go as far as applying command pattern though.

damithc: We can leave it as an exercise too.

src/seedu/addressbook/Main.java
```diff
@@ -0,0 +1,394 @@
...
+ * This class header comment below is brief because details of how to
+ * use this class are documented elsewhere.
+ * ====================================================================
+ */
+
+/**
+ * Entry point of the address book application.
+ * Contains command execution logic; initialises and uses the different components.
+ */
+public class Main implements Runnable {
```
damithc: Can we do without the `Runnable` interface?

src/seedu/addressbook/TextUi.java
```diff
@@ -0,0 +1,327 @@
...
+     * Ignores empty, pure whitespace, and comment lines.
+     *
+     * @see Parser#shouldIgnore(String)
+     * @return full line entered by the user
+     */
+    public String getUserCommand() {
+        out.print(LINE_PREFIX + "Enter command: ");
+        String fullInputLine = in.nextLine();
+        // silently consume all ignored lines
+        while (parser.shouldIgnore(fullInputLine)) {
```
damithc: This line seems to be the only reason for TextUi to know the Parser. See if you can get rid of it. You can move the `shouldIgnore()` method to the TextUi itself.

damithc:
Seems generally OK. Can go ahead and merge. Pending stuff can be done in separate PRs. 
We probably need a `Command` <|-- `AddCommand` hierarchy before we can claim this code is OOP. But we can do it after we reach the full version, and only if we have time. I need a lot of time to weave in the exercises, so we need to reach the full version soon.

damithc:
BTW, there is no need to do unit tests for the entire code base. You can omit some less important areas of code untested. Just leave empty test classes/methods  with `TODO: implement this` inside.

src/seedu/addressbook/TextUi.java
YijinL: I was thinking of it. Just the part where we upgrade commands into first class objects, but we can ignore invokers etc. I would like to do it, it will shorten the code and reduce complexity by quite a bit.

src/seedu/addressbook/Utils.java
YijinL: In terms of modularity, each data type class (`Person`, `Name`, `UniqueTagList`) etc has certain constraints they must fulfill. 

So if we see a random `Name` object, we know that it conforms to the naming format, if we see a random `UniqueTagList` object, we know that it contains no duplicates etc..

Is there some nicer way of working in a not-null guarantee? I'd really like instances of data classes to behave as expected/specified in every possible case.

src/seedu/addressbook/TextUi.java
YijinL: So it will be a half-version of the command pattern, like how we already have a half-versino of the MVC pattern here.

src/seedu/addressbook/model/person/UniquePersonList.java
YijinL: I agree; however when we move to javaFX, the view uses `ObservableList` and having to propogate changes in the set to the obs-list negates most of the time complexity advantages of using a set. 

So should we keep it as a list? (easier migration to javafx based level 3)
or use a set? (semantically better, but has to propogate changes to list view)

PR #9, opened 2016-08-08
Convert to semi-command pattern #6

### LINKED ISSUE #6: Use command objects to reduce clutter spread out across components
[old discussion here](https://github.com/se-edu/addressbook-level2/pull/1#discussion_r73797588)

YijinL:
Resolves #6.

Points to note
- `Parser` removed; command level parsing now handled in the command objects
- `Main`  now does preliminary parsing (split command word and command args, multiplex correct command object)
- Even with >10  new class files (with all the imports and headers etc), the resulting source is smaller in terms of LoC, how nice.


damithc:
Good work. I would keep the Parser class though. Parsing the input is not Command's job. Parser is the object that takes a string and returns a Command object.

src/seedu/addressbook/commands/Command.java
```diff
@@ -0,0 +1,16 @@
+package seedu.addressbook.commands;
+
+/**
+ *
+ */
+@FunctionalInterface
```
damithc: Don't use functional programming stuff.

damithc:
Can merge after addressing comments.

YijinL:
@damithc 
So when students want to extend it with a new command, say `EditPerson`, then they have to create a new command object AND write more lines in the parser class? Then one of the advantages (easy to extend) of command objects is lost

YijinL:
Then if we want to extend functionality with a new command, we have to
1. write the new command class
2. add more methods to a common bloated parser class 
Wouldn't it be better to wrap everything command-specific, from parsing to execution, in the same object?

damithc:
Not really. The logic to build an object, if it is complex, is often put outside the object. Hence the Factory pattern. The parser is in fact a factory class of sorts. 
Adding a new command requires a change to the command language, which requires a change to the parser.

YijinL:
Oh i see... so the parser class will be quite big then.
Would it make sense to have separate parser objects for each command :P

damithc:
Let's do one parser class first. We can always break it up later if we need to, without affecting anything else.

YijinL:
Will merge this first and have parser extraction in a separate PR

PR #14, opened 2016-08-14
Minor cleanups to Main class and Command classes

damithc:
* reduced unnecessary coupling to TextUi 
* made Main class simpler


src/seedu/addressbook/commands/HelpCommand.java
```diff
@@ -1,29 +1,30 @@
...
+        return new CommandResult(
+                AddPersonCommand.MESSAGE_USAGE
+                + "\n" + DeletePersonCommand.MESSAGE_USAGE
+                + "\n" + ClearAddressBookCommand.MESSAGE_USAGE
+                + "\n" + FindPersonsByWordsInNameCommand.MESSAGE_USAGE
+                + "\n" + ListAllPersonsCommand.MESSAGE_USAGE
+                + "\n" + ViewPersonDetailsCommand.MESSAGE_USAGE
+                + "\n" + ViewAllPersonDetailsCommand.MESSAGE_USAGE
+                + "\n" + HelpCommand.MESSAGE_USAGE
+                + "\n" + ExitCommand.MESSAGE_USAGE
```
damithc: `\n` is replaced by the correct formatting before displaying to the user. Formatting should only happen in the TextUi.

src/seedu/addressbook/commands/IncorrectCommand.java
```diff
@@ -0,0 +1,20 @@
...
+    String feedbackToUser;
+
+    public IncorrectCommand(String feedbackToUser){
+        this.feedbackToUser = feedbackToUser;
+    }
+
+    @Override
+    public CommandResult execute() {
+        return new CommandResult(feedbackToUser);
+    }
```
damithc: It's better to user a normal object rather than an exception to handle user mistakes. They are not truly exceptional situations.

damithc:
@YijinL for your review and merging.

damithc:
Basically, I was fixing one small problem at a time (as you can see from the commits) and running the tests are each fix.

src/seedu/addressbook/Main.java
```diff
@@ -1,160 +1,140 @@
...
+     * Sets up the different components, loads up the data from the storage file, and prints the welcome message.
      *
-     * @param storageFilePath file path of the desired storage file
+     * @param launchArgs arguments supplied by the user at program launch
      * @param inputStream user text input source
      * @param outputStream user text output acceptor
      *
-     * @throws MainInitialisationException if there were problems
      */
-    public Main(String storageFilePath, InputStream inputStream, PrintStream outputStream)
```
damithc: Removed init code elsewhere because It is recommended that the constructor is kept as simple as possible.

src/seedu/addressbook/Main.java
```diff
@@ -1,160 +1,140 @@
...
+     * Reads the user command and executes it, until the user issues the exit command.
      */
-    private static String getStorageFilePathFromLaunchArgs(String... launchArgs) {
-        return launchArgs.length > 0 ? launchArgs[0] : DEFAULT_STORAGE_FILEPATH;
-    }
+    private void runCommandLoopUntilExitCommand() {
+        Command command;
+        do {
+            String userCommand = ui.getUserCommand();
+            command = new Parser().parseCommand(userCommand);
```
damithc: Moved parsing here. Parsing should not be inside execution.

src/seedu/addressbook/Main.java
```diff
@@ -1,160 +1,140 @@
...
+    private TextUi ui;
+    private StorageFile storage;
+    private AddressBook addressBook;
 
     /**
-     * Version info of the program.
+     * The list of person shown to the user most recently.
      */
-    public static final String VERSION = "AddessBook Level 1 - Version 1.0";
+    private List<? extends ReadOnlyPerson> lastShownList = null;
```
damithc: I wonder if this is the right datatype.  Feels like it is too low level. We can encapsulate it into another class at a later stage (not very important)

src/seedu/addressbook/commands/Command.java
```diff
@@ -1,24 +1,42 @@
...
 import seedu.addressbook.model.AddressBook;
+import seedu.addressbook.model.person.ReadOnlyPerson;
+
+import java.util.List;
 
 /**
  * Represents a command with hidden internal logic and the ability to be executed.
  */
-public interface Command {
+public abstract class Command {
```
damithc: Converted to a class because that seems like the better choice. But I'll think of how to remove the `abstract` as it is not covered yet.

src/seedu/addressbook/Main.java
```diff
@@ -1,160 +1,140 @@
...
-        ui.showWelcomeMessage(VERSION);
-        ui.showToUser(String.format(MESSAGE_USING_STORAGE_FILE, storageFile.toString()));
-        // [read input, execute, response] loop
-        while (true) {
-            String userCommand = ui.getUserCommand();
-            ui.echoLastEnteredUserCommand();
-            String feedback = executeCommand(userCommand);
-            ui.showResultToUser(feedback);
+    private void updateLastShownList(CommandResult result) {
+        if(result.getRelevantPersons() != null) {
```
YijinL: Would it make sense to use an Optional here? This `CommandResult.relevantPersons` is the only place in the code where we use nulls.

src/seedu/addressbook/ui/TextUi.java
```diff
@@ -98,71 +73,63 @@ public boolean isCommentLine(String rawInputLine) {
...
     }
 
     /**
      * Shows the result of a command execution to the user. Includes additional formatting to demarcate different
      * command execution segments.
      */
-    public void showResultToUser(String result) {
-        showToUser(result, DIVIDER);
+    public void showResultToUser(CommandResult result) {
+        if(result.getRelevantPersons() != null) {showPersonListView(result.getRelevantPersons());}
```
YijinL: Same here; only place with purposeful nulls

src/seedu/addressbook/ui/TextUi.java
```diff
@@ -98,71 +73,63 @@ public boolean isCommentLine(String rawInputLine) {
...
+        showToUser(MESSAGE_INIT_FAILED, DIVIDER, DIVIDER);
     }
 
     /**
      * Shows message(s) to the user
      */
     public void showToUser(String... message) {
         for (String m : message) {
-            out.println(LINE_PREFIX + m);
+            out.println(LINE_PREFIX + m.replace("\n", LS + LINE_PREFIX));
```
YijinL: The regex `\R` will match any unicode line break sequence.

src/seedu/addressbook/Main.java
damithc: Yes, this is a good place for `Optional`, but let's not introduce it in this version. That kind of 'optional' topics can be delayed. In early levels we should limit ourselves to essentials only.

src/seedu/addressbook/ui/TextUi.java
damithc: I want to limit use of regex to the parser only. Even within parser I would use  it only in some places, just to contrast with the simpler approach. regex is not easy to learn and not essential at early stages.

test/expected.txt
```diff
@@ -289,6 +289,8 @@ Launch arguments format: [STORAGE_FILE_PATH]
 || 0 persons listed!
 || ===================================================
 || Enter command: || [Command entered:  exit]
-|| Exiting Address Book... Good bye!
+|| Exiting Addressbook as requested ...
```
damithc: Should be `Address Book`.

damithc:
See my comment about not using regex.

YijinL:
Fixed. Will merge.

PR #22, opened 2016-08-20
More minor cleanups

damithc:
Renaming, revising comments etc.


src/seedu/addressbook/Main.java
```diff
@@ -1,65 +1,56 @@
 package seedu.addressbook;
 
-import seedu.addressbook.model.person.ReadOnlyPerson;
+import seedu.addressbook.data.person.ReadOnlyPerson;
```
damithc: We haven't introduced the 'model-view' stuff yet.

src/seedu/addressbook/Main.java
```diff
@@ -1,65 +1,56 @@
...
-    /**
-     * The list of person shown to the user most recently.
-     */
+    /** The list of person shown to the user most recently.  */
     private List<? extends ReadOnlyPerson> lastShownList = null;
 
 
     public static void main(String... launchArgs) {
-        new Main().run(launchArgs, System.in, System.out);
+        new Main().run(launchArgs);
```
damithc: I guess you added these two parameters to use Junit for I/O testing? If yes, we'll bring them back later when we need it.

src/seedu/addressbook/Main.java
```diff
@@ -1,65 +1,56 @@
...
+    /** Runs the program until termination.  */
+    public void run(String[] launchArgs) {
+        start(launchArgs);
         runCommandLoopUntilExitCommand();
         exit();
     }
 
     /**
-     * Sets up the different components, loads up the data from the storage file, and prints the welcome message.
+     * Sets up the required objects, loads up the data from the storage file, and prints the welcome message.
```
damithc: Trying not to use 'components' as we have not come to that stage yet. As the design is supposed to be flat, there are no components.

src/seedu/addressbook/data/person/ReadOnlyPerson.java
```diff
@@ -21,7 +21,7 @@
     UniqueTagList getTags();
 
     /**
-     * Convenience method for comparing equality via state. (interfaces cannot override .equals)
+     * Returns true if the values inside this object is same as those of the other (Note: interfaces cannot override .equals)
```
damithc: For boolean methods, always use this phrasing (`Returns true ...`).

src/seedu/addressbook/data/person/UniquePersonList.java
```diff
@@ -33,46 +31,27 @@ protected DuplicatePersonException() {
...
-        internalList.addAll(persons);
-    }
-
-    /**
-     * Copy constructor, insulates from changes in source.
+     * Constructs a shallow copy of the list.
      */
     public UniquePersonList(UniquePersonList source) {
-        internalList.addAll(source.internalList); // insulate internal list from changes in argument
+        internalList.addAll(source.internalList);
```
damithc: So the constructor can access private variables of another object of the same class? Seems odd :-p TIL

src/seedu/addressbook/data/tag/UniqueTagList.java
```diff
@@ -49,37 +47,27 @@ public UniqueTagList(Tag... tags) throws DuplicateTagException {
...
+     * Constructs a tag list with the given tags.
      */
     public UniqueTagList(Set<Tag> tags) {
-        Utils.assertNoNullElements(tags);
         internalList.addAll(tags);
     }
 
     /**
-     * Copy constructor, insulates from changes in source.
+     * Constructs a shallow copy of the given tag list.
```
damithc: It is a shallow copy, right?

damithc:
@YijinL for review and merging.

src/seedu/addressbook/data/tag/UniqueTagList.java
YijinL: Yes, though `Tag` objects are immutable for now.

src/seedu/addressbook/Main.java
YijinL: Yes, ok we put them in when we need them

src/seedu/addressbook/data/person/UniquePersonList.java
YijinL: Yeah since it's in the class scope, it has full access.

src/seedu/addressbook/ui/TextUi.java
```diff
@@ -18,65 +17,62 @@
...
 
-    /**
-     * A platform independent line separator.
-     */
-    public static final String LS = System.lineSeparator();
+    /** A platform independent line separator. */
+    private static final String LS = System.lineSeparator();
 
-    public static final String DIVIDER = "===================================================";
+    private static final String DIVIDER = "===================================================";
```
YijinL: I'm actually curious why we make it private, when it's a final immutable object there's no security concerns right?

damithc: It's not just security. We should not expose a member unless it's needed by an external party. A bit like having glass walls in a home - we want to use them sparingly even if they are unbreakable glass.

PR #25, opened 2016-08-20
Move model creation outside parser #20

### LINKED ISSUE #20: Move person/tag creation code out of the parser
Parser should not be creating persons/tags. The AddCommand should contain the values given by the user rather than a Person object.

YijinL:
Fixes #20


damithc:
:+1:  Can merge

PR #29, opened 2016-08-26
Add runtests.sh

jaeoheeail:
I translated runtests.bat file so I could test it on a Mac

### DIFF
--- a/test/runtests.sh
+++ b/test/runtests.sh
@@ -1,5 +1,4 @@
-#!/bin/bash
-# runtests.sh
+#!/usr/bin/env bash
 
 # create bin directory if it doesn't exist
 if [ ! -d "../bin" ]
@@ -8,7 +7,7 @@ then
 fi
 
 # delete output from previous run
-if [ ! -d "../actual.txt"]
+if [ ! -d "../actual.txt" ]
 then
     rm actual.txt
 fi
--- a/test/runtests.sh
+++ b/test/runtests.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# runtests.sh
+
+# create bin directory if it doesn't exist
+if [ ! -d "../bin" ]
+then 
+	mkdir ../bin
+fi
+
+
+# delete output from previous run
+rm actual.txt
+
+# compile the code into the bin folder
+javac -cp ../src -Xlint:none -d ../bin ../src/seedu/addressbook/Main.java
+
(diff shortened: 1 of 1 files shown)


damithc:
@ndt93 can test this and merge?

test/runtests.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# runtests.sh
+
+# create bin directory if it doesn't exist
+if [ ! -d "../bin" ]
+then
```
ndt93: Can remove trailing whitespace on this line?

test/runtests.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# runtests.sh
+
+# create bin directory if it doesn't exist
+if [ ! -d "../bin" ]
+then 
+	mkdir ../bin
```
ndt93: Maybe use `spaces` instead of `tab` here to be consistent with the `if` block below.

test/runtests.sh
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+
+# delete output from previous run
+rm actual.txt
```
ndt93: You might want to check if the file exists here. Also 1 blank line above should be enough I suppose.

ndt93:
@jaeoheeail The script runs fine. I added some comments on cosmetic issues. I will merge this after those are fixed. Thanks!

test/runtests.sh
```diff
@@ -0,0 +1,29 @@
...
+# runtests.sh
+
+# create bin directory if it doesn't exist
+if [ ! -d "../bin" ]
+then
+    mkdir ../bin
+fi
+
+# delete output from previous run
+if [ ! -d "../actual.txt"]
```
ndt93: There should be a space between the last `"` and `]`. `Bash` is smart enough to figure this out but there will be a warning because it's not the proper syntax.

ndt93:
@jaeoheeail Added one more comment. Can you also commit the file with the execution permission bit set at least for the user since I'm not sure if some students would be confused when they can't execute the file.

test/runtests.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# runtests.sh
```
ndt93: No need to put the file name here? :-P

jaeoheeail: @ndt93 Sorry for the extra space, fixed it and changed the permission

PR #30, opened 2016-08-26
[T3A3][T14-A2]Tan Min Rong Samson

samsontmr:
Ready for Review.


damithc:
This PR should be against [nus-cs2103-AY1617S1](https://github.com/nus-cs2103-AY1617S1/addressbook-level1) :-)
Closing...

PR #36, opened 2016-08-26
[35] NullPointerException if viewall is used before list

### LINKED ISSUE #35: NullPointerException if viewall is used before list
The "viewall" command, produces a Null Pointer Exception, even when there is a person at given index. To replicate the issue add a few users and close the program. Then restart and try the first command as viewall 1. However, if we try the list command before viewall the issue disappears. From the code it turns out that viewall works on the last shown list and not the stored file. I am not sure if this was intended or not.

YijinL:
Fixes #35.

### DIFF
--- a/src/seedu/addressbook/Main.java
+++ b/src/seedu/addressbook/Main.java
@@ -9,6 +9,7 @@ import seedu.addressbook.parser.Parser;
 import seedu.addressbook.storage.StorageFile;
 import seedu.addressbook.ui.TextUi;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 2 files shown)


damithc:
1. Update tests?
2. Rename PR to match normal convention

damithc:
Can merge after that.

YijinL:
`runtests.sh` passes.

damithc:
I meant you should add a test to cover this situation.

damithc:
You are also confusing commit message format with PR name format. 
PR name should be a copy paste of issue name with the trailing (not leading) issue number.
In future we can consider making them both the same, but for now they are different.

YijinL:
Oops fixed my other PR names too.
Regression test added in 1ed45b79f8ba48919029686b83aa63a1bcc12603

PR #94, opened 2016-09-09
[T4A3][F11-B2]ChiangWengKiat

wengkiat:
Created parent class contact and removed common methods in child classes


wengkiat:
Ready for review

Issue #108, opened 2016-11-27
Add unit tests for AddCommand class


damithc:
as per title

PierceAndy:
Issue closed by #118.

PR #43, opened 2016-08-30
Name regex doesn't match hyphen, dots and quote #41

okkhoy:
Fixes #41

PR #118, opened 2016-12-08
Add unit tests for AddCommand class #108

Neurrone:
Ready for review.

### DIFF
--- a/src/seedu/addressbook/data/person/Address.java
+++ b/src/seedu/addressbook/data/person/Address.java
@@ -21,11 +21,12 @@ public class Address {
      * @throws IllegalValueException if given address string is invalid.
      */
     public Address(String address, boolean isPrivate) throws IllegalValueException {
+        String trimmedAddress = address.trim();
         this.isPrivate = isPrivate;
-        if (!isValidAddress(address)) {
+        if (!isValidAddress(trimmedAddress)) {
             throw new IllegalValueException(MESSAGE_ADDRESS_CONSTRAINTS);
         }
-        this.value = address;
+        this.value = trimmedAddress;
     }
 
     /**
--- a/src/seedu/addressbook/data/person/Name.java
+++ b/src/seedu/addressbook/data/person/Name.java
@@ -12,7 +12,7 @@ import java.util.List;
 public class Name {
 
     public static final String EXAMPLE = "John Doe";
-    public static final String MESSAGE_NAME_CONSTRAINTS = "Person names should be spaces or alphabetic characters";
+    public static final String MESSAGE_NAME_CONSTRAINTS = "Names should consist of letters, spaces, numbers, ', - or .";
     public static final String NAME_VALIDATION_REGEX = "^[\\p{L}0-9 .'-]+$";  
     // In the above:
     // \\p{L} matches any unicode letter (like German name with accents)
--- a/test/java/seedu/addressbook/commands/AddCommandTest.java
+++ b/test/java/seedu/addressbook/commands/AddCommandTest.java
@@ -101,7 +101,7 @@ public class AddCommandTest {
                 Address.EXAMPLE, true, EMPTY_STRING_LIST);
         ReadOnlyPerson p = command.getPerson();
 
-        // todo: add comparison of tags to person.equals and equality methods to
+        // TODO: add comparison of tags to person.equals and equality methods to
         // individual fields that compare privacy to simplify this
         assertEquals(Name.EXAMPLE, p.getName().fullName);
         assertEquals(Phone.EXAMPLE, p.getPhone().value);
--- a/test/java/seedu/addressbook/commands/AddCommandTest.java
+++ b/test/java/seedu/addressbook/commands/AddCommandTest.java
@@ -1,0 +1,147 @@
+package seedu.addressbook.commands;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.junit.Test;
+
+import seedu.addressbook.data.AddressBook;
+import seedu.addressbook.data.exception.IllegalValueException;
+import seedu.addressbook.data.person.Address;
+import seedu.addressbook.data.person.Email;
+import seedu.addressbook.data.person.Name;
+import seedu.addressbook.data.person.Person;
+import seedu.addressbook.data.person.Phone;
+import seedu.addressbook.data.person.ReadOnlyPerson;
+import seedu.addressbook.data.person.UniquePersonList;
+import seedu.addressbook.util.TestUtil;
+
+public class AddCommandTest {
+    private static final List<ReadOnlyPerson> EMPTY_PERSON_LIST = Collections.emptyList();
+    private static final Set<String> EMPTY_STRING_LIST = Collections.emptySet();
+
+    @Test
+    public void addCommand_invalidName_throwsException() {
+        final String[] invalidNames = { "", " ", "[]\\[;]" };
+        for (String name : invalidNames) {
+            assertConstructingInvalidAddCmdThrowsException(name, Phone.EXAMPLE, true, Email.EXAMPLE, false,
+                    Address.EXAMPLE, true, EMPTY_STRING_LIST);
+        }
+    }
+
+    @Test
+    public void addCommand_invalidPhone_throws() {
+        final String[] invalidNumbers = { "", " ", "1234-5678", "[]\\[;]", "abc", "a123", "+651234" };
+        for (String number : invalidNumbers) {
+            assertConstructingInvalidAddCmdThrowsException(Name.EXAMPLE, number, false, Email.EXAMPLE, true,
+                    Address.EXAMPLE, false, EMPTY_STRING_LIST);
+        }
+    }
+
+    @Test
+    public void addCommand_invalidEmail_throws() {
+        final String[] invalidEmails = { "", " ", "def.com", "@", "@def", "@def.com", "abc@",
+                "@invalid@email", "invalid@email!", "!invalid@email" };
+        for (String email : invalidEmails) {
+            assertConstructingInvalidAddCmdThrowsException(Name.EXAMPLE, Phone.EXAMPLE, false, email, false,
+                    Address.EXAMPLE, false, EMPTY_STRING_LIST);
+        }
+    }
+
+    @Test
+    public void addCommand_invalidAddress_throws() {
+        final String[] invalidAddresses = { "", " " };
+        for (String address : invalidAddresses) {
+            assertConstructingInvalidAddCmdThrowsException(Name.EXAMPLE, Phone.EXAMPLE, true, Email.EXAMPLE,
+                    true, address, true, EMPTY_STRING_LIST);
+        }
+    }
+
+    @Test
+    public void addCommand_invalidTags_throws() {
+        final String[][] invalidTags = { { "" }, { " " }, { "'" }, { "[]\\[;]" }, { "validTag", "" },
+                { "", " " } };
+        for (String[] tags : invalidTags) {
+            Set<String> tagsToAdd = new HashSet<>(Arrays.asList(tags));
+            assertConstructingInvalidAddCmdThrowsException(Name.EXAMPLE, Phone.EXAMPLE, true, Email.EXAMPLE,
+                    true, Address.EXAMPLE, false, tagsToAdd);
+        }
+    }
+
+    /**
+     * Asserts that attempting to construct an add command with the supplied
+     * invalid data throws an IllegalValueException
+     */
+    private void assertConstructingInvalidAddCmdThrowsException(String name, String phone,
+            boolean isPhonePrivate, String email, boolean isEmailPrivate, String address,
+            boolean isAddressPrivate, Set<String> tags) {
+        try {
+            new AddCommand(name, phone, isPhonePrivate, email, isEmailPrivate, address, isAddressPrivate,
+                    tags);
+        } catch (IllegalValueException e) {
+            return;
+        }
+        String error = String.format(
+                "An add command was successfully constructed with invalid input: %s %s %s %s %s %s %s %s",
+                name, phone, isPhonePrivate, email, isEmailPrivate, address, isAddressPrivate, tags);
+        fail(error);
+    }
+
+    @Test
+    public void addCommand_validData_correctlyConstructed() throws Exception {
+        AddCommand command = new AddCommand(Name.EXAMPLE, Phone.EXAMPLE, true, Email.EXAMPLE, false,
+                Address.EXAMPLE, true, EMPTY_STRING_LIST);
+        ReadOnlyPerson p = command.getPerson();
+
+        // todo: add comparison of tags to person.equals and equality methods to
+        // individual fields that compare privacy to simplify this
+        assertEquals(Name.EXAMPLE, p.getName().fullName);
+        assertEquals(Phone.EXAMPLE, p.getPhone().value);
+        assertTrue(p.getPhone().isPrivate());
+        assertEquals(Email.EXAMPLE, p.getEmail().value);
+        assertFalse(p.getEmail().isPrivate());
+        assertEquals(Address.EXAMPLE, p.getAddress().value);
+        assertTrue(p.getAddress().isPrivate());
+        boolean isTagListEmpty = !p.getTags().iterator().hasNext();
+        assertTrue(isTagListEmpty);
+    }
+
+    @Test
+    public void addCommand_emptyAddressBook_addressBookContainsPerson() {
+        Person p = TestUtil.generateTestPerson();
+        AddCommand command = new AddCommand(p);
+        AddressBook book = new AddressBook();
+        command.setData(book, EMPTY_PERSON_LIST);
+        CommandResult result = command.execute();
+        UniquePersonList people = book.getAllPersons();
+
+        assertTrue(people.contains(p));
+        assertEquals(1, people.immutableListView().size());
+        assertFalse(result.getRelevantPersons().isPresent());
+        assertEquals(String.format(AddCommand.MESSAGE_SUCCESS, p), result.feedbackToUser);
+    }
+
+    @Test
+    public void addCommand_addressBookAlreadyContainsPerson_addressBookUnmodified() throws Exception {
+        Person p = TestUtil.generateTestPerson();
+        AddressBook book = new AddressBook();
+        book.addPerson(p);
+        AddCommand command = new AddCommand(p);
+        command.setData(book, EMPTY_PERSON_LIST);
+        CommandResult result = command.execute();
+
+        assertFalse(result.getRelevantPersons().isPresent());
+        assertEquals(AddCommand.MESSAGE_DUPLICATE_PERSON, result.feedbackToUser);
+        UniquePersonList people = book.getAllPersons();
+        assertTrue(people.contains(p));
+        assertEquals(1, people.immutableListView().size());
+    }
+}
--- a/test/java/seedu/addressbook/commands/AddCommandTest.java
+++ b/test/java/seedu/addressbook/commands/AddCommandTest.java
@@ -38,7 +38,7 @@ public class AddCommandTest {
     }
 
     @Test
-    public void addCommand_invalidPhone_throws() {
+    public void addCommand_invalidPhone_throwsException() {
         final String[] invalidNumbers = { "", " ", "1234-5678", "[]\\[;]", "abc", "a123", "+651234" };
         for (String number : invalidNumbers) {
             assertConstructingInvalidAddCmdThrowsException(Name.EXAMPLE, number, false, Email.EXAMPLE, true,
@@ -47,7 +47,7 @@ public class AddCommandTest {
     }
 
     @Test
-    public void addCommand_invalidEmail_throws() {
+    public void addCommand_invalidEmail_throwsException() {
         final String[] invalidEmails = { "", " ", "def.com", "@", "@def", "@def.com", "abc@",
                 "@invalid@email", "invalid@email!", "!invalid@email" };
         for (String email : invalidEmails) {
@@ -57,7 +57,7 @@ public class AddCommandTest {
     }
 
     @Test
-    public void addCommand_invalidAddress_throws() {
+    public void addCommand_invalidAddress_throwsException() {
         final String[] invalidAddresses = { "", " " };
         for (String address : invalidAddresses) {
             assertConstructingInvalidAddCmdThrowsException(Name.EXAMPLE, Phone.EXAMPLE, true, Email.EXAMPLE,
@@ -66,7 +66,7 @@ public class AddCommandTest {
     }
 
     @Test
-    public void addCommand_invalidTags_throws() {
+    public void addCommand_invalidTags_throwsException() {
         final String[][] invalidTags = { { "" }, { " " }, { "'" }, { "[]\\[;]" }, { "validTag", "" },
                 { "", " " } };
         for (String[] tags : invalidTags) {
--- a/test/java/seedu/addressbook/util/TestUtil.java
+++ b/test/java/seedu/addressbook/util/TestUtil.java
@@ -1,12 +1,20 @@
 package seedu.addressbook.util;
 
+import static org.junit.Assert.fail;
+
 import java.util.ArrayList;
 import java.util.List;
 
 import seedu.addressbook.data.AddressBook;
+import seedu.addressbook.data.exception.IllegalValueException;
+import seedu.addressbook.data.person.Address;
+import seedu.addressbook.data.person.Email;
+import seedu.addressbook.data.person.Name;
 import seedu.addressbook.data.person.Person;
+import seedu.addressbook.data.person.Phone;
 import seedu.addressbook.data.person.ReadOnlyPerson;
 import seedu.addressbook.data.person.UniquePersonList.DuplicatePersonException;
+import seedu.addressbook.data.tag.UniqueTagList;
 
 public class TestUtil {
     /**
@@ -42,5 +50,14 @@ public class TestUtil {
      */
     public static AddressBook clone(AddressBook addressBook) {
         return new AddressBook(addressBook.getAllPersons(), addressBook.getAllTags());
+    }
+    public static Person generateTestPerson() {
+        try {
+            return new Person(new Name(Name.EXAMPLE), new Phone(Phone.EXAMPLE, false),
+                    new Email(Email.EXAMPLE, true), new Address(Address.EXAMPLE, false), new UniqueTagList());
+        } catch (IllegalValueException e) {
+            fail("test person data should be valid by definition");
+            return null;
+        }
     }
 }


damithc:
@pyokagan can review this one? I tried assigning you as a reviewer, but didn't work (may be a UI bug)

damithc:
@Neurrone I assume you are an SoC student. Have you taken CS2103 yet?

Neurrone:
Oh, just realized I haven't filled out my github profile. Yeah, Dickson here.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,192 @@
...
+import seedu.addressbook.data.person.UniquePersonList.DuplicatePersonException;
+import seedu.addressbook.data.tag.UniqueTagList;
+
+public class AddCommandTest {
+	private static final List<? extends ReadOnlyPerson> LAST_SHOWN_LIST = Collections.emptyList();
+	
+	/*
+	 * Note how the names of the test methods does not follow the normal naming
+	 * convention. That is because our coding standard allows a different naming
+	 * convention for test methods.
```
pyokagan: Actually, not sure if @damithc will want this comment repeated for every test :-P

Neurrone:
The test failures are caused by "'" being accepted as a valid name, and " " as a valid address. Do we want to test for these conditions?

Will work on fixing.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,192 @@
...
+		
+		AddCommand cmd = new AddCommand(p);
+		cmd.setData(book, LAST_SHOWN_LIST);
+		CommandResult res = cmd.execute();
+		
+		assertFalse(res.getRelevantPersons().isPresent());
+		assertEquals(res.feedbackToUser, AddCommand.MESSAGE_DUPLICATE_PERSON);
+		UniquePersonList people = book.getAllPersons();
+		assertTrue(people.contains(p));
+		assertTrue(countPeople(people) == 1);
```
Neurrone: ?

test/java/seedu/addressbook/commands/AddCommandTest.java
damithc: @Neurrone can remove this comment. No need to repeat it in every class. In fact, we should move it out of the code and mention it once in a more central location, but that's a separate issue.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,192 @@
...
+			ConstructingInvalidAddCmdThrows(Name.EXAMPLE, Phone.EXAMPLE, false, Email.EXAMPLE, false, Address.EXAMPLE,
+					false, tagsToAdd);
+		}
+	}
+
+	/**
+	 * Asserts that attempting to construct an add command with the supplied
+	 * invalid data throws an IllegalValueException
+	 */
+	private void ConstructingInvalidAddCmdThrows(String name, String phone, boolean isPhonePrivate, String email,
```
damithc: We usually name this kind of methods with `assert...` to indicate that the method has an assert-like behavior inside it. e.g `assertConstructingInvalidAddCommandThrowsException`

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,192 @@
...
+			boolean isEmailPrivate, String address, boolean isAddressPrivate, Set<String> tags) {
+		try {
+			new AddCommand(name, phone, isPhonePrivate, email, isEmailPrivate, address,
+					isAddressPrivate, tags);
+			final String errorFormatter = new String(
+					"An add command was successfully constructed with invalid input: %s %s %s %s %s %s %s %s");
```
damithc: Yes, I think Codacy flagged this too.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,192 @@
...
+			assertEquals(p.getEmail().value, Email.EXAMPLE);
+			assertFalse(p.getEmail().isPrivate());
+			assertEquals(p.getAddress().value, Address.EXAMPLE);
+			assertTrue(p.getAddress().isPrivate());
+			boolean isTagListEmpty = !p.getTags().iterator().hasNext();
+			assertTrue(isTagListEmpty);
+		} 
+		catch (IllegalValueException e) {
+			fail("Adding command with valid data failed.");
+		}
```
damithc: I think the try/catch can be omitted. Just let the exception go through (i.e. add a throws clause to the method signature). If the exception happens, it will result in a test failure anyway.

test/java/seedu/addressbook/commands/AddCommandTest.java
pyokagan: Ah, it's just a small nit, but I was wondering if
```java
assertTrue(countPeople(people) == 1);
```
could be expressed as
```java
assertEquals(1, countPeople(people));
```

Neurrone: Got it.

test/java/seedu/addressbook/commands/AddCommandTest.java
pyokagan: @damithc 
> (i.e. add a throws clause to the method signature)

Hmm, would adding a `throws Exception` be okay? Or do you prefer the `throws` clause to be as specific as possible (e.g. `throws IllegalValueException`)

pyokagan: Reason I ask is because during my project when I introduced a new checked exception it was hard work updating all of the `throws` clauses. :-P

pyokagan:
@Neurrone 
> The test failures are caused by "'" being accepted as a valid name, and " " as a valid address. Do we want to test for these conditions?

For `Name` accepting `'`, it seems that it was purposefully introduced in #43 so I guess someone has a use case for that. So maybe we should not test that as an invalid input.

I also noticed that even though the regex was updated in #43 , the `MESSAGE_NAME_CONSTRAINTS` message was not :-P. Perhaps this could be fixed in your PR.

For Address accepting ` `, it seems that the `Address` constructor [does not actually trim its input](https://github.com/se-edu/addressbook-level2/blob/7518552f786121d505f2ae1613901eae745d631e/src/seedu/addressbook/data/person/Address.java#L23-L29), as opposed to the other `Name`, `Phone`, `Email` classes. So yeah, it's behaviour is inconsistent with the other `Person` fields, so perhaps this should be considered a bug that can be fixed in your PR.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+    private static Person generateTestPerson() {
+        try {
+            return new Person(new Name(Name.EXAMPLE), new Phone(Phone.EXAMPLE, false),
+                    new Email(Email.EXAMPLE, true), new Address(Address.EXAMPLE, false), new UniqueTagList());
+        } catch (IllegalValueException e) {
+            fail("test person data should be valid by definition");
+            return null;
+        }
+    }
+}
```
pyokagan: @Neurrone Oh, one small thing -- can add a newline at the end of file? For consistency with the other source files in this repo.

Neurrone:
Oops, the latest commit was an attempted rebase to fix a typo gone wrong. I guess all these individual commits will be squashed into one once done?

pyokagan:
@Neurrone 
> I guess all these individual commits will be squashed into one once done?

Yeah, they will. Still, if you are interested, you can use `git reflog` and `git reset` to undo your botched rebase. (This is just FYI, don't need to do it)

e.g. if your branch name is `108-add-command-unit-tests` you can just run
```
git reflog 108-add-command-unit-tests
```

Git will then show you the history of SHA1s of the head of that branch. You can the just copy the SHA1 you want to reset the branch back to, and then run

```
git reset --hard THE_SHA1
```

Neurrone:
@pyokagan thanks for the tip.

src/seedu/addressbook/data/person/Name.java
```diff
@@ -6,26 +6,28 @@
...
+ * Represents a Person's name in the address book. Guarantees: immutable; is
+ * valid as declared in {@link #isValidName(String)}
  */
 public class Name {
-
+    
     public static final String EXAMPLE = "John Doe";
-    public static final String MESSAGE_NAME_CONSTRAINTS = "Person names should be spaces or alphabetic characters";
-    public static final String NAME_VALIDATION_REGEX = "^[\\p{L}0-9 .'-]+$";  
+    public static final String MESSAGE_NAME_CONSTRAINTS = "Person names should consist of unicode letters, spaces, numbers, ', - or .";
```
pyokagan: This message is user-facing, so probably remove the `unicode` since that's a technical term that users don't need to know about.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+import seedu.addressbook.data.person.UniquePersonList.DuplicatePersonException;
+import seedu.addressbook.data.tag.UniqueTagList;
+
+public class AddCommandTest {
+    private static final List<ReadOnlyPerson> LAST_SHOWN_LIST = Collections.emptyList();
+    
+    @Test
+    public void addCommand_invalidName_throws() {
+        Set<String> tags = new HashSet<>();
+        final String[] invalidNames = { "", " ", "[]\\[;]" };
```
pyokagan: So we test at least one character, at least one character (with space trimming), and any invalid character. Looks good.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+import seedu.addressbook.data.person.UniquePersonList;
+import seedu.addressbook.data.person.UniquePersonList.DuplicatePersonException;
+import seedu.addressbook.data.tag.UniqueTagList;
+
+public class AddCommandTest {
+    private static final List<ReadOnlyPerson> LAST_SHOWN_LIST = Collections.emptyList();
+    
+    @Test
+    public void addCommand_invalidName_throws() {
+        Set<String> tags = new HashSet<>();
```
pyokagan: Maybe instead of constructing an empty `HashSet` every time for every test, maybe define a global `NO_TAGS` constant or something that is assigned the value of `Collections.emptySet()`.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+        for (String name : invalidNames) {
+            assertConstructingInvalidAddCmdThrows(name, Phone.EXAMPLE, false, Email.EXAMPLE, false,
+                    Address.EXAMPLE, false, tags);
+        }
+    }
+    
+    @Test
+    public void addCommand_invalidPhone_throws() {
+        Set<String> tags = new HashSet<>();
+        final String[] invalidNumbers = { "", " ", "1234-5678", "[]\\[;]", "abc", "a123", "+651234" };
```
pyokagan: OK.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+        for (String number : invalidNumbers) {
+            assertConstructingInvalidAddCmdThrows(Name.EXAMPLE, number, false, Email.EXAMPLE, false,
+                    Address.EXAMPLE, false, tags);
+        }
+    }
+    
+    @Test
+    public void addCommand_invalidEmail_throws() {
+        Set<String> tags = new HashSet<>();
+        final String[] invalidEmails = { "", " ", "def.com", "@", "@def", "@def.com", "abc@" };
```
pyokagan: OK, these test cases adequately test the invalid inputs where the right side of the `@` is empty and the left side of the `@` is empty.

In addition, maybe test for characters that we reject on the right side and left side of the `@`?

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+        for (String email : invalidEmails) {
+            assertConstructingInvalidAddCmdThrows(Name.EXAMPLE, Phone.EXAMPLE, false, email, false,
+                    Address.EXAMPLE, false, tags);
+        }
+    }
+    
+    @Test
+    public void addCommand_invalidAddress_throws() {
+        Set<String> tags = new HashSet<>();
+        final String[] invalidAddresses = { "", " " };
```
pyokagan: OK

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+        for (String address : invalidAddresses) {
+            assertConstructingInvalidAddCmdThrows(Name.EXAMPLE, Phone.EXAMPLE, false, Email.EXAMPLE, false,
+                    address, false, tags);
+        }
+    }
+    
+    @Test
+    public void addCommand_invalidTags_throws() {
+        final String[][] invalidTags = { { "" }, { " " }, { "'" }, { "validTag", "" }, { "", " " } };
```
pyokagan: Maybe also test for characters that are not accepted as a tag name e.g. `[]\\[;]`

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+        for (String[] tags : invalidTags) {
+            Set<String> tagsToAdd = new HashSet<>(Arrays.asList(tags));
+            assertConstructingInvalidAddCmdThrows(Name.EXAMPLE, Phone.EXAMPLE, false, Email.EXAMPLE, false,
+                    Address.EXAMPLE, false, tagsToAdd);
+        }
+    }
+
```
pyokagan: Make sure all empty lines are actually empty. (This line, and the rest of the lines in this file, all contain spaces)

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+        AddCommand cmd = new AddCommand(Name.EXAMPLE, Phone.EXAMPLE, true, Email.EXAMPLE, false,
+                Address.EXAMPLE, true, tags); // should never throw
+        ReadOnlyPerson p = cmd.getPerson();
+        
+        assertEquals(p.getName().fullName, Name.EXAMPLE);
+        assertEquals(p.getPhone().value, Phone.EXAMPLE);
+        assertTrue(p.getPhone().isPrivate());
+        assertEquals(p.getEmail().value, Email.EXAMPLE);
+        assertFalse(p.getEmail().isPrivate());
+        assertEquals(p.getAddress().value, Address.EXAMPLE);
```
pyokagan: Take note of the argument order of [`assertEquals`](http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(java.lang.Object,%20java.lang.Object)) -- The first argument is the expected value and the second argument is the actual value.

src/seedu/addressbook/data/person/Address.java
```diff
@@ -21,11 +21,12 @@
      * @throws IllegalValueException if given address string is invalid.
      */
     public Address(String address, boolean isPrivate) throws IllegalValueException {
+        String addr = address.trim();
```
damithc: `trimmedAddress`? We try to avoid partial words and short forms.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+                "An add command was successfully constructed with invalid input: %s %s %s %s %s %s %s %s",
+                name, phone, isPhonePrivate, email, isEmailPrivate, address, isAddressPrivate, tags);
+        fail(error);
+    }
+
+    @Test
+    public void addCommand_validData_correctlyConstructed() throws Exception {
+        Set<String> tags = new HashSet<>();
+        AddCommand cmd = new AddCommand(Name.EXAMPLE, Phone.EXAMPLE, true, Email.EXAMPLE, false,
+                Address.EXAMPLE, true, tags); // should never throw
```
pyokagan: Can use `NO_TAGS` here as well?

Neurrone: Oops, will fix. Thanks for all the feedback :)

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,155 @@
...
+        for (String[] tags : invalidTags) {
+            Set<String> tagsToAdd = new HashSet<>(Arrays.asList(tags));
+            assertConstructingInvalidAddCmdThrows(Name.EXAMPLE, Phone.EXAMPLE, false, Email.EXAMPLE, false,
+                    Address.EXAMPLE, false, tagsToAdd);
+        }
+    }
+
+    /**
+     * Asserts that attempting to construct an add command with the supplied
+     * invalid data throws an exception
```
pyokagan: I think @damithc was referring to the case where we *do not expect an exception*.

However, in this test, we are testing specifically for an exception (IllegalValueException), so I think we need to test specifically for it.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,155 @@
...
+     */
+    private void assertConstructingInvalidAddCmdThrows(String name, String phone, boolean isPhonePrivate,
+            String email, boolean isEmailPrivate, String address, boolean isAddressPrivate,
+            Set<String> tags) {
+        try {
+            new AddCommand(name, phone, isPhonePrivate, email, isEmailPrivate, address, isAddressPrivate,
+                    tags);
+        } catch (Exception e) {
```
pyokagan: So I think this should be changed back to `IllegalValueException`.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,155 @@
...
+        UniquePersonList people = book.getAllPersons();
+        assertTrue(people.contains(p));
+        assertEquals(1, people.immutableListView().size());
+    }
+
+    private static Person generateTestPerson() {
+        try {
+            return new Person(new Name(Name.EXAMPLE), new Phone(Phone.EXAMPLE, false),
+                    new Email(Email.EXAMPLE, true), new Address(Address.EXAMPLE, false), new UniqueTagList());
+        } catch (Exception e) {
```
pyokagan: Same for here, I think this should be reverted back to `IllegalValueException`.

test/java/seedu/addressbook/commands/AddCommandTest.java
pyokagan: Yeah, `Exception` will actually catch *all* exceptions, which is probably what we do not want here.

test/java/seedu/addressbook/commands/AddCommandTest.java
Neurrone: Sorry, misunderstood the exception conversation earlier as 'wanting to not specify a specific exception' vs 'don't need to specify an exception if we expect none to be thrown'

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+import seedu.addressbook.data.person.Email;
+import seedu.addressbook.data.person.Name;
+import seedu.addressbook.data.person.Person;
+import seedu.addressbook.data.person.Phone;
+import seedu.addressbook.data.person.ReadOnlyPerson;
+import seedu.addressbook.data.person.UniquePersonList;
+import seedu.addressbook.data.tag.UniqueTagList;
+
+public class AddCommandTest {
+    private static final List<ReadOnlyPerson> LAST_SHOWN_LIST = Collections.emptyList();
```
damithc: Rename to `EMPTY_PERSON_LIST`?

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+    private static final Set<String> NO_TAGS = Collections.emptySet();
```
damithc: Rename to `EMPTY_STRING_LIST`?

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Test
+    public void addCommand_invalidName_throws() {
```
damithc: `_throwsException()` (to be consistent with other places; I don't think we use `throws` as a short form of `throwsException` in any other place)

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+        final String[] invalidNames = { "", " ", "[]\\[;]" };
+        for (String name : invalidNames) {
+            assertConstructingInvalidAddCmdThrows(name, Phone.EXAMPLE, false, Email.EXAMPLE, false,
+                    Address.EXAMPLE, false, NO_TAGS);
```
damithc: Seems we always use `false` as the test input. It's better to mix it up and use `true` in some places. Increasing the variety of test inputs can increase the chance of finding unexpected bugs.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+        String error = String.format(
+                "An add command was successfully constructed with invalid input: %s %s %s %s %s %s %s %s",
+                name, phone, isPhonePrivate, email, isEmailPrivate, address, isAddressPrivate, tags);
+        fail(error);
+    }
+
+    @Test
+    public void addCommand_validData_correctlyConstructed() throws Exception {
+        AddCommand cmd = new AddCommand(Name.EXAMPLE, Phone.EXAMPLE, true, Email.EXAMPLE, false,
+                Address.EXAMPLE, true, NO_TAGS); // should never throw
```
damithc: Comment can be removed. Not throwing an exception is the normal expectation that goes without saying.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+        assertEquals(Name.EXAMPLE, p.getName().fullName);
+        assertEquals(Phone.EXAMPLE, p.getPhone().value);
+        assertTrue(p.getPhone().isPrivate());
+        assertEquals(Email.EXAMPLE, p.getEmail().value);
+        assertFalse(p.getEmail().isPrivate());
+        assertEquals(Address.EXAMPLE, p.getAddress().value);
+        assertTrue(p.getAddress().isPrivate());
+        boolean isTagListEmpty = !p.getTags().iterator().hasNext();
+        assertTrue(isTagListEmpty);
```
damithc: Any chance of using Person class `equals` method to do this comparison?

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+    private static Person generateTestPerson() {
+        try {
+            return new Person(new Name(Name.EXAMPLE), new Phone(Phone.EXAMPLE, false),
+                    new Email(Email.EXAMPLE, true), new Address(Address.EXAMPLE, false), new UniqueTagList());
+        } catch (IllegalValueException e) {
+            fail("test person data should be valid by definition");
+            return null;
+        }
+    }
```
damithc: Perhaps we can move this to a separate test util class so that it can be shared by other test classes? Test classes often contain duplicate utility methods which is something we should try to avoid.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+        assertFalse(res.getRelevantPersons().isPresent());
+        assertEquals(String.format(AddCommand.MESSAGE_SUCCESS, p), res.feedbackToUser);
+    }
+
+    @Test
+    public void addCommand_addressBookAlreadyContainsPerson_addressBookUnmodified()
+            throws Exception {
+        Person p = generateTestPerson();
+        AddressBook book = new AddressBook();
+        book.addPerson(p); // should never throw
```
damithc: Comment can be removed.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+        assertTrue(isTagListEmpty);
+    }
+
+    @Test
+    public void addCommand_emptyAddressBook_addressBookContainsPerson() {
+        Person p = generateTestPerson();
+        AddCommand cmd = new AddCommand(p);
+        AddressBook book = new AddressBook();
+        cmd.setData(book, LAST_SHOWN_LIST);
+        CommandResult res = cmd.execute();
```
damithc: `res` --> `result`

test/java/seedu/addressbook/commands/AddCommandTest.java
Neurrone: Yeah. I noticed that that method doesn't compare tag equality. Is that intentional?

damithc: That could be simply a 'shortcut' taken by the developer to save time (it was done in a rush :-).
Anyway, I think Person needs multiple equality methods each with a different level of strictness.
Sometimes we want to check if both objects contain the same exact values (I think that's the one you want), and other times we just want to check if it 'is the same person' (i.e. just compare the identity-related fields such as the name).  You can do this bit as a separate PR. Just remember to add an issue (and a `TODO` in code) so that we don't forget about it.

test/java/seedu/addressbook/commands/AddCommandTest.java
Neurrone: Looks like this overlaps with https://github.com/se-edu/addressbook-level2/pull/120 Should I wait till this is merged before using the TestUtils stuff there?

damithc: This PR is probably nearer to completion than that one, so no need to wait. Whichever PR that merges later will have to retrofit its code to match the one merged earlier :-)

test/java/seedu/addressbook/commands/AddCommandTest.java
Neurrone: cmd to command as well?

damithc:
@Neurrone A TestUtil class has been added by the recent PR #119

Neurrone:
Have integrated latest changes and squashed.

damithc:
@Neurrone A few more places where `throws` should be changed to `throwsException`. Can merge after that.

test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -0,0 +1,147 @@
...
+        fail(error);
+    }
+
+    @Test
+    public void addCommand_validData_correctlyConstructed() throws Exception {
+        AddCommand command = new AddCommand(Name.EXAMPLE, Phone.EXAMPLE, true, Email.EXAMPLE, false,
+                Address.EXAMPLE, true, EMPTY_STRING_LIST);
+        ReadOnlyPerson p = command.getPerson();
+
+        // todo: add comparison of tags to person.equals and equality methods to
```
pyokagan: @Neurrone Minor nit: this should be `TODO` in capital letters.

pyokagan: Mainly because I think developers are more likely to do `git grep TODO` rather than `git grep todo` or `git grep -i TODO`.

Also, some editors apply special highlighting to `TODO` and not `todo`.

pyokagan: Other than this everything looks fine.

PR #120, opened 2016-12-09
Add unit tests for FindCommand class #110

KnewYouWereTrouble:
Fixes #110
Ready for review.

---
This change is [Reviewable](https://reviewable.io/reviews/se-edu/addressbook-level2/120)

PR #121, opened 2016-12-09
Add unit tests for AddressBook class #105

### LINKED ISSUE #105: Add unit tests for AddressBook class
as per title

thenaesh:
Fixes #105
Ready for review

---
This change is [Reviewable](https://reviewable.io/reviews/se-edu/addressbook-level2/121)


test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,359 @@
...
+		personBill   = new Person(new Name("Bill"),
+                                   new Phone("154982351", false),
+                                   new Email("bill@whitehouse.gov", false),
+                                   new Address("White House", false),
+                                   new UniqueTagList(tagDemocrat));
+		personMonroe = new Person(new Name("Monroe"),
+                                   new Phone("168463354", false),
+                                   new Email("email@notinvented.yet", false),
+                                   new Address("James Monroe Tomb, Richmond, Virginia", false),
+                                   new UniqueTagList(tagIndependent, tagDead));
```
damithc: When defining test data, try to follow a pattern that is easy to remember. For example, the first person's name has initials A. B. (e.g. Alice Betsy), second person's name has initials B. C. (Ben Charles) and so on. When things always follow a pattern, anything going out of that pattern (e.g. Ben suddenly appears at the start of the list) gets noticed immediately and sometimes can lead to discovering bugs.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,359 @@
...
+import seedu.addressbook.data.tag.UniqueTagList.TagNotFoundException;
+
+/**
+ * we check, in particular, for the two guarantees specified by the AddressBook class
+ *  - Every tag found in every person will also be found in the tag list.
+ *  - The tags in each person point to tag objects in the master list. (== equality)
+ */
+public class AddressBookTest {
+	
+	private Tag tagDead;
```
damithc: Wrong indentation. Should be 4 spaces.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,359 @@
...
+		UniqueTagList tags = new UniqueTagList(tagDemocrat, tagRepublican);
+		UniquePersonList persons = new UniquePersonList(personBarack, personBill);
+		
+		AddressBook addressBook = new AddressBook(persons, tags);
+		
+		return addressBook;
+	}
+	
+	/**
+	 * checks if the underlying container behind an iterable is empty
```
damithc: `Checks`

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,359 @@
...
+import seedu.addressbook.data.person.UniquePersonList;
+import seedu.addressbook.data.person.UniquePersonList.DuplicatePersonException;
+import seedu.addressbook.data.person.UniquePersonList.PersonNotFoundException;
+import seedu.addressbook.data.tag.Tag;
+import seedu.addressbook.data.tag.UniqueTagList;
+import seedu.addressbook.data.tag.UniqueTagList.DuplicateTagException;
+import seedu.addressbook.data.tag.UniqueTagList.TagNotFoundException;
+
+/**
+ * we check, in particular, for the two guarantees specified by the AddressBook class
```
damithc: `We`

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,359 @@
...
+	
+	private Tag tagDead;
+	private Tag tagRepublican;
+	private Tag tagDemocrat;
+	private Tag tagIndependent;
+
+	private Person personBarack; // has tagDemocrat
+	private Person personGeorge; // has tagRepublican
+	private Person personBill;   // has tagDemocrat
+	private Person personMonroe; // has tagIndependent, tagDead
```
damithc: OSS projects are supposed to last long and cross culture boundaries. So it is best not to pick names/values that may not be known to all or may be known as well after some time.

damithc: In other words, code should not give subtle hints as to the period in which it was written or the person who wrote it.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,359 @@
...
+import seedu.addressbook.data.person.UniquePersonList.PersonNotFoundException;
+import seedu.addressbook.data.tag.Tag;
+import seedu.addressbook.data.tag.UniqueTagList;
+import seedu.addressbook.data.tag.UniqueTagList.DuplicateTagException;
+import seedu.addressbook.data.tag.UniqueTagList.TagNotFoundException;
+
+/**
+ * we check, in particular, for the two guarantees specified by the AddressBook class
+ *  - Every tag found in every person will also be found in the tag list.
+ *  - The tags in each person point to tag objects in the master list. (== equality)
```
damithc: By convention this class tests the AddressBook class. So there is no need to explain what functionality it tests. It should test _all_ functionality of the AddressBook class. Tests can be organized based on public methods of the class such that for each public method there is one or more test methods.

thenaesh:
@damithc I've made changes taking into account all the comments mentioned. I've removed the redundant comment at the start of the test class, fixed the indentation and some other formatting. I've also made the test cases culturally-neutral and also made them follow the pattern suggested.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,372 @@
...
+    public void addTag_addNonDuplicate_addsNormally() throws Exception {
+        AddressBook addressBook = createDefaultAddressBook();
+
+        assertFalse(addressBook.containsTag(tagEconomist));
+        addressBook.addTag(tagEconomist);
+        assertTrue(addressBook.containsTag(tagEconomist));
+    }
+
+    @Test
+    public void addTag_addDuplicate_throwsDuplicateTagException() throws Exception {
```
damithc: `addTag_tagAlreadyInList_throwsDuplicateTagException`

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    @Test
+    public void containsPerson_personNotExists_returnsFalse() throws Exception {
+        AddressBook addressBook = createDefaultAddressBook();
+
+        boolean foundPerson = addressBook.containsPerson(personCharlieDouglas);
+
+        assertFalse(foundPerson);
+    }
```
damithc: Can be shorter
```java
@Test
public void containsPerson() throws Exception {
    AddressBook addressBook = createDefaultAddressBook();
    assertTrue(addressBook.containsPerson(personAliceBetsy));
    assertFalse(addressBook.containsPerson(personCharlieDouglas));
}
```

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,372 @@
...
+        // Tag constructor will not throw as the strings supplied match the
+        // regex \p{Alnum}+
+        tagPrizeWinner   = new Tag("prizewinner");
+        tagScientist     = new Tag("scientist");
+        tagMathematician = new Tag("mathematician");
+        tagEconomist     = new Tag("economist");
+
+        // UniqueTagList constructor will not throw as there are no duplicate
+        // elements
+        personAliceBetsy     = new Person(new Name("Alice Betsy"),
```
damithc: `person` prefix can be omitted.

thenaesh:
@damithc I've moved the helper methods out into the TestUtils class and cleaned up the unnecessary comments. I've also combined, renamed and simplified tests according to the suggestions.

I've modified the tests expecting exceptions to be thrown to use JUnit's ExpectedException instead of manually catching the exceptions. As PR #120 has not been merged yet, I am not using PersonBuilder to create the Person objects at the moment.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+    private Person aliceBetsy; // has tagMathematician
+    private Person bobChaplin; // has tagMathematician
+    private Person charlieDouglas; // has tagScientist
+    private Person davidElliot; // has tagEconomist, tagPrizeWinner
+
+    /**
+     * Initialises an AddressBook with some default values.
+     * 
+     * Contains personAliceBetsy and personBobChaplin.
```
damithc: `person` prefix should be deleted.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,275 @@
...
+    @Test
+    public void syncTagsWithMasterList_emptyInitialTagList_addsAllIntoTagList() throws Exception {
+        UniqueTagList emptyTagList = new UniqueTagList();
+        UniquePersonList persons = new UniquePersonList(bobChaplin, charlieDouglas);
+
+        AddressBook addressBook = new AddressBook(persons, emptyTagList);
+
+        assertTrue(addressBook.containsTag(tagMathematician));
+        assertTrue(addressBook.containsTag(tagScientist));
+        assertFalse(addressBook.containsTag(tagEconomist));
```
damithc: Can this be done in one shot (and a more definitive way)? e.g. 
`assertEquals(expectedTagList, addressBook.getAllTags())`

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,275 @@
...
+            // particular person tag is equal to some tag in the master list
+            boolean personTagValid = false;
+            for (Tag tag : tags) {
+                personTagValid = personTagValid || (personTag == tag);
+            }
+            allPersonTagsValid = allPersonTagsValid && personTagValid;
+        }
+        
+        assertTrue(allPersonTagsValid);
+    }
```
damithc: What does this test case tests that the previous methods don't?
I think the only functionality we can test is when a person is added, if that person has a tag that is missing in the tag list, it will be added to the list. In fact, it can be tested while testing `addPerson` method.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,275 @@
...
+        addressBook.addPerson(davidElliot);
+        assertTrue(addressBook.containsPerson(davidElliot));
+    }
+
+    @Test
+    public void addPerson_personAlreadyInList_throwsDuplicatePersonException() throws Exception {
+        AddressBook addressBook = createDefaultAddressBook();
+        
+        thrown.expect(DuplicatePersonException.class);
+        addressBook.addPerson(aliceBetsy);
```
damithc: Here, you can verify that if the duplicate person has a tag that is not in the list, it will not be added to the list.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    @Test
+    public void addTag_tagNotInList_addsNormally() throws Exception {
+        AddressBook addressBook = createDefaultAddressBook();
```
damithc: Looks like this line appears in every `@Test`. May be put that in the `@Before` method?

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,275 @@
...
+        
+        thrown.expect(DuplicateTagException.class);
+        addressBook.addTag(tagMathematician);
+    }
+
+    @Test
+    public void containsPerson() throws Exception {
+        AddressBook addressBook = createDefaultAddressBook();
+
+        assertTrue(addressBook.containsPerson(aliceBetsy));
```
damithc: Loop over all Persons that is supposed to be in the list?
You should also test this method for empty list.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,275 @@
...
+        assertTrue(addressBook.containsTag(tagScientist));
+        assertFalse(addressBook.containsTag(tagEconomist));
+    }
+
+    @Test
+    public void removePerson_personExists_removesNormally() throws Exception {
+        AddressBook addressBook = createDefaultAddressBook();
+
+        assertTrue(addressBook.containsPerson(aliceBetsy));
+        addressBook.removePerson(aliceBetsy);
```
damithc: Should at least check the size to ensure no more than the target person was removed.

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -51,6 +56,34 @@ public static AddressBook createAddressBook(Person... persons) throws DuplicateP
     public static AddressBook clone(AddressBook addressBook) {
         return new AddressBook(addressBook.getAllPersons(), addressBook.getAllTags());
     }
+
+    /**
+     * Returns true iff every pair of corresponding elements two iterables are (deeply) identical.
```
damithc: The method is order-sensitive? If yes, should mention in the header comment.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,275 @@
...
+public class AddressBookTest {
+    private Tag tagPrizeWinner;
+    private Tag tagScientist;
+    private Tag tagMathematician;
+    private Tag tagEconomist;
+
+    private Person aliceBetsy; // has tagMathematician
+    private Person bobChaplin; // has tagMathematician
+    private Person charlieDouglas; // has tagScientist
+    private Person davidElliot; // has tagEconomist, tagPrizeWinner
```
damithc: The info in these comments are also present in another place (in the `setUp` method), and therefore this is a violation of the DRY principle. Comments can be removed.

thenaesh:
@damithc I've made all the changes requested, including spreading out the tag equality tests.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,289 @@
...
+     * is equal (as references) to some tag in the master list of that AddressBook.
+     */
+    private boolean checkTagEqualityProperty(Tag tagToCheck, AddressBook addressBook) {
+        boolean tagEqualityPropertyHolds = false;
+        
+        for (Tag tag : addressBook.getAllTags()) {
+            tagEqualityPropertyHolds = tagEqualityPropertyHolds || (tag == tagToCheck);
+        }
+        
+        return tagEqualityPropertyHolds;
```
damithc: Alternative algo (shorter):
```java
for (Tag tag : addressBook.getAllTags()) {
    if (tag == tagToCheck) { return true; }
}

return false;
```

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,289 @@
...
+                                    new Address("10 Science Drive", false),
+                                    new UniqueTagList(tagScientist));
+
+        davidElliot    = new Person(new Name("David Elliot"),
+                                    new Phone("84512575", false),
+                                    new Email("douglas@nuscomputing.com", false),
+                                    new Address("11 Arts Link", false),
+                                    new UniqueTagList(tagEconomist, tagPrizeWinner));
+        
+        defaultAddressBook = createDefaultAddressBook();
```
damithc: Use this instead?
```java
defaultAddressBook = new AddressBook(new UniquePersonList(aliceBetsy, bobChaplin),
                                     new UniqueTagList(tagMathematician, tagScientist));

File: src/seedu/addressbook/data/exception/ConstraintViolationException.java [hunk 21]
```diff
@@ -0,0 +1,13 @@
+package seedu.addressbook.data.exception;
+
+/**
+ * Signals that the operation should not take place as it will violate some constraint.
+ */
+public class ConstraintViolationException extends Exception {
```
damithc: Why are you introducing a new exception just for testing?

thenaesh:
@damithc I've made all the changes requested, except for the `ConstraintViolationException`. I'm quite confused with the implementation of `removeTag` in the `AddressBook` class. The guarantees specified in comments at the top of the file state that "every tag found in every person will also be found in the tag list", but `removeTag` will happily delete any tag as long as it can find it in the tag list, even if there is still a person with that tag in the address book.

I'm not sure if I'm misunderstanding the meaning of the guarantee, or if this behaviour is actually a bug. If it is the former, perhaps you could clarify and I can then correct the test case. If it is the latter, and if the abovementioned scenario arises, shouldn't `removeTag` throw an exception that is something other than `TagNotFoundException`? I introduced a new exception because this seemed to be the case. If I shouldn't introduce a new exception, then should I just test for `Exception`, or do something else entirely?

damithc:
@thenaesh that sounds like a bug. In that case, it may be ok to throw the exception.  Go ahead.

In the meantime, @YijinL any input from you?

thenaesh:
I've added a commit fixing that bug.

There's also another failing test `addPerson_personAlreadyInListButHasTagNotInList_tagNotAdded`, which fails because `addPerson` does a `syncTagsWithMasterList` before attempting to actually add the `Person` (which may throw). I've also added another commit fixing that bug (which is simply swapping the two lines around).

I can revert the commits if the fixes are not desired (or if I should perform them in a separate PR).

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,253 @@
...
+    @Test
+    public void getAllTags() throws Exception {
+        UniqueTagList allTags = defaultAddressBook.getAllTags();
+        UniqueTagList tagsToCheck = new UniqueTagList(tagMathematician, tagScientist);
+
+        assertTrue(isIdentical(allTags, tagsToCheck));
+    }
+    
+    
+    // Helper Methods
```
damithc: No need

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -51,6 +56,48 @@ public static AddressBook createAddressBook(Person... persons) throws DuplicateP
...
+                return false;
+            }
+        }
+
+        // If any of the two iterables still have elements, then they have different sizes.
+        return !(currentPtr0.hasNext() || currentPtr1.hasNext());
+    }
+    
+    /**
+     * Returns true iff the underlying container behind an iterable is empty.
```
damithc: Do we need the `iff`? In other places we use `if` even when we mean `if and only if`

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -51,6 +56,48 @@ public static AddressBook createAddressBook(Person... persons) throws DuplicateP
...
+     */
+    public static <T> int getSize(Iterable<T> it) {
+        int numberOfElementsSeen = 0;
+        
+        for (T elem : it) {
+            numberOfElementsSeen++;
+        }
+        
+        return numberOfElementsSeen;
+    }
```
damithc: I'm a bit sceptical that we have to write all these utility methods for iterators. Google a bit to see if there are standard solutions.

src/seedu/addressbook/data/AddressBook.java
```diff
@@ -118,7 +119,12 @@ public void removePerson(ReadOnlyPerson toRemove) throws PersonNotFoundException
      *
      * @throws TagNotFoundException if no such Tag could be found.
      */
-    public void removeTag(Tag toRemove) throws TagNotFoundException {
+    public void removeTag(Tag toRemove) throws TagNotFoundException, ConstraintViolationException {
+        for (Person person : allPersons) {
+            if (person.getTags().contains(toRemove)) {
+                throw new ConstraintViolationException("The Tag to be removed is present in a Person object.");
+            }
+        }
```
damithc: Extract a `isTagUsed()` method to improve SLAP?

YijinL:
Yes, `removeTag` should check and remove the appropriate tag from all persons in the addressbook.

damithc:
> Yes, removeTag should check and remove the appropriate tag from all persons in the addressbook.
@thenaesh that sounds a better approach than the current approach. In that case we can do without the exception too.

test/java/seedu/addressbook/util/TestUtil.java
thenaesh: It seems that there are no standard solutions that do not involve adding new dependencies (e.g. Guava).

thenaesh:
@damithc I've made the requested changes, except for replacing the helper methods with standard solutions (as I couldn't find any that didn't involve adding new dependencies).

damithc:
@thenaesh this branch is outdated. Can rebase?

thenaesh:
@damithc Oops, sorry forgot. Done.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,258 @@
...
+        aliceBetsy.setTags(new UniqueTagList(tagMathematician, tagPrizeWinner));
+        
+        try {
+            defaultAddressBook.addPerson(aliceBetsy);
+        } catch (DuplicatePersonException e) {
+            assertFalse(defaultAddressBook.containsTag(tagPrizeWinner));
+            return;
+        }
+        
+        fail("Expected DuplicatePersonException was not caught.");
```
damithc: This flow more intuitive?

```java
try {
     defaultAddressBook.addPerson(aliceBetsy);
     fail("Expected DuplicatePersonException was not thrown.");
} catch (DuplicatePersonException e) {
   // ignore expected exception
}

assertFalse(defaultAddressBook.containsTag(tagPrizeWinner));
       
```

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,258 @@
...
+    public void containsTag() throws Exception {
+        assertTrue(defaultAddressBook.containsTag(tagScientist));
+        assertFalse(defaultAddressBook.containsTag(tagEconomist));
+        
+        UniqueTagList allTags = new UniqueTagList(tagPrizeWinner, tagScientist, tagMathematician, tagEconomist);
+        
+        for (Tag tag : allTags) {
+            assertFalse(emptyAddressBook.containsTag(tag));
+        }
+    }
```
damithc: Why the above two methods have different structures? They are testing similar things right?

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+    @Test
+    public void removePerson_personExists_removesNormally() throws Exception {
+        int numberOfPersonsBeforeRemoval = getSize(defaultAddressBook.getAllPersons());
+        defaultAddressBook.removePerson(aliceBetsy);
+        int numberOfPersonsAfterRemoval = getSize(defaultAddressBook.getAllPersons());
```
damithc: This line can be moved further down. Don't define a variable until you need it.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,258 @@
...
+        thrown.expect(TagNotFoundException.class);
+        defaultAddressBook.removeTag(tagEconomist);
+    }
+
+    @Test
+    public void clear() throws Exception {
+        defaultAddressBook.clear();
+
+        // If the iterator does not have next element at start, then underlying
+        // container has no elements.
```
damithc: Sounds like this comment belongs to the `isEmpty()` method.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,258 @@
...
+    @Test
+    public void getAllTags() throws Exception {
+        UniqueTagList allTags = defaultAddressBook.getAllTags();
+        UniqueTagList tagsToCheck = new UniqueTagList(tagMathematician, tagScientist);
+
+        assertTrue(isIdentical(allTags, tagsToCheck));
+    }
+    
+    /**
+     * Returns true if the given Tag object if found in the tag list of the given AddressBook.
```
damithc: `if found` -> `is found`?

damithc:
@thenaesh reminder

thenaesh:
@damithc Sorry for the delay, I've made the changes requested.

test/java/seedu/addressbook/data/AddressBookTest.java
damithc: Resolved

test/java/seedu/addressbook/data/AddressBookTest.java
damithc: Resolved

test/java/seedu/addressbook/data/AddressBookTest.java
damithc: Resolved

test/java/seedu/addressbook/data/AddressBookTest.java
damithc: Resolved

src/seedu/addressbook/data/AddressBook.java
```diff
@@ -119,8 +119,21 @@ public void removePerson(ReadOnlyPerson toRemove) throws PersonNotFoundException
      * @throws TagNotFoundException if no such Tag could be found.
      */
     public void removeTag(Tag toRemove) throws TagNotFoundException {
+        removeTagFromAllPersons(toRemove);
```
damithc: The header comment can be updated to mention this behavior.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -0,0 +1,265 @@
...
+        
+        int numberOfTagsBeforeRemoval = getSize(defaultAddressBook.getAllTags());
+        defaultAddressBook.removeTag(tagMathematician);
+        
+        assertTrue(isIdentical(aliceBetsy.getTags(), new UniqueTagList(tagPrizeWinner)));
+        assertTrue(isIdentical(bobChaplin.getTags(), new UniqueTagList(tagEconomist)));
+        assertFalse(defaultAddressBook.containsTag(tagMathematician));
+        
+        int numberOfTagsAfterRemoval = getSize(defaultAddressBook.getAllTags());
+        assertTrue(numberOfTagsBeforeRemoval == numberOfTagsAfterRemoval + 1);
```
damithc: Is this more intuitive? `numberOfTagsAfterRemoval == numberOfTagsBeforeRemoval - 1`

pyokagan:
@damithc Sorry, but with the semester starting I don't have any more capacity for immediate reviews.

damithc:
> with the semester starting I don't have any more capacity for immediate reviews.

Noted @pyokagan. No worries.

thenaesh:
@damithc I've made the requested changes.

PR #127, opened 2016-12-17
Add unit tests for Utils class #107

### LINKED ISSUE #107: Add unit tests for Utils class
as per title

zzzzwen:
Fixes #107
Ready for review

### DIFF
--- a/test/java/seedu/addressbook/common/UtilsTest.java
+++ b/test/java/seedu/addressbook/common/UtilsTest.java
@@ -11,28 +11,29 @@ import org.junit.Test;
 public class UtilsTest {
     @Test
     public void isAnyNull() {
-        List<Object> emptyList = createList();
-        List<Object> nonEmptyList = createList("A string");
-
         // empty list
         assertFalse(Utils.isAnyNull());
 
         // Any non-empty list
         assertFalse(Utils.isAnyNull(new Object(), new Object()));
-        assertFalse(Utils.isAnyNull(emptyList));
+        assertFalse(Utils.isAnyNull("test"));
 
         // non empty list with just one at the beginning
         assertTrue(Utils.isAnyNull((Object)null));
         assertTrue(Utils.isAnyNull(null, "", new Object(), null));
-        assertTrue(Utils.isAnyNull(null, emptyList, nonEmptyList));
+        assertTrue(Utils.isAnyNull(null, new Object(), new Object()));
 
         // non empty list with nulls in the middle
-        assertTrue(Utils.isAnyNull(nonEmptyList, null, null, emptyList));
+        assertTrue(Utils.isAnyNull(new Object(), null, null, "test"));
         assertTrue(Utils.isAnyNull("", null, new Object()));
 
         // non empty list with one null as the last element
         assertTrue(Utils.isAnyNull("", new Object(), null));
-        assertTrue(Utils.isAnyNull(nonEmptyList, emptyList, null));
+        assertTrue(Utils.isAnyNull(new Object(), new Integer(1), null));
+        
+        // test whether method peeks into the elements inside container
+        List<Object> emptyList = createList((Object)null);
+        assertFalse(Utils.isAnyNull(emptyList));
     }
 
     @Test
--- a/test/java/seedu/addressbook/common/UtilsTest.java
+++ b/test/java/seedu/addressbook/common/UtilsTest.java
@@ -4,6 +4,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 
 import org.junit.Before;
 import org.junit.Test;
@@ -23,7 +24,7 @@ public class UtilsTest {
         tag1Copy = buildTag("tag1");
         tag2 = buildTag("tag2");
     }
-    
+
     private static Tag buildTag(String tagName) {
         try {
             return new Tag(tagName);
@@ -34,10 +35,10 @@ public class UtilsTest {
 
     @Test
     public void isAnyNull_noNulls_returnsFalse() {
-        //empty list
+        // empty list
         assertFalse(Utils.isAnyNull());
-        
-        //Any non-empty list
+
+        // Any non-empty list
         ArrayList<String> emptyList = new ArrayList<String>();
         assertFalse(Utils.isAnyNull(new Object(), new Object()));
         assertFalse(Utils.isAnyNull(emptyList));
@@ -48,107 +49,46 @@ public class UtilsTest {
         ArrayList<String> emptyList = new ArrayList<String>();
         ArrayList<String> nonEmptyList = new ArrayList<String>();
         nonEmptyList.add("A string");
-        
-        //non empty list with just one at the beginning
+
+        // non empty list with just one at the beginning
         assertTrue(Utils.isAnyNull(nullObject, "", new Object()));
         assertTrue(Utils.isAnyNull(nullObject, emptyList, nonEmptyList));
-        
-        //non empty list with nulls in the middle
+
+        // non empty list with nulls in the middle
         assertTrue(Utils.isAnyNull(nonEmptyList, nullObject, nullObject, emptyList));
         assertTrue(Utils.isAnyNull("", nullObject, new Object()));
-        
-        //non empty list with one null as the last element
+
+        // non empty list with one null as the last element
         assertTrue(Utils.isAnyNull("", new Object(), nullObject));
         assertTrue(Utils.isAnyNull(nonEmptyList, emptyList, nullObject));
     }
 
     @Test
-    public void elementsAreUnique_emptyCollection_returnsTrue() {
-        assertTrue(Utils.elementsAreUnique(new ArrayList<>()));
+    public void elementsAreUnique() {
+        //empty list
+        assertAreUnique();
+
+        //only one object
+        assertAreUnique(nullObject);
+        assertAreUnique(tag1);
+
+        //all objects unique
+        assertAreUnique("abc", "ab", "a");
+        assertAreUnique(tag1, tag2);
+        
+        //some identical objects
+        assertNotUnique("abc", "", "abc");
+        assertNotUnique(tag1, tag1Copy, tag2);
     }
 
-    @Test
-    public void elementsAreUnique_singleElement_returnsTrue() {
-        assertTrue(Utils.elementsAreUnique(new ArrayList<Object>() {
-            {
-                add(nullObject);
-            }
-        }));
-        assertTrue(Utils.elementsAreUnique(new ArrayList<String>() {
-            {
-                add("String1");
-            }
-        }));
-        assertTrue(Utils.elementsAreUnique(new ArrayList<Tag>() {
-            {
-                add(tag1);
-            }
-        }));
+    void assertAreUnique(Object... objects) {
+        ArrayList<Object> list = new ArrayList<Object>(Arrays.asList(objects));
+        assertTrue(Utils.elementsAreUnique(list));
     }
 
-    @Test
-    public void elementsAreUnique_sameElements_returnsFalse() {
-        assertFalse(Utils.elementsAreUnique(new ArrayList<String>() {
-            {
-                add("abc");
-                add("abc");
-            }
-        }));
-        assertFalse(Utils.elementsAreUnique(new ArrayList<Tag>() {
-            {
-                add(tag1);
-                add(tag1);
-                add(tag1);
-            }
-        }));
-    }
-
-    @Test
-    public void elementsAreUnique_allNulls_returnsFalse() {
-        assertFalse(Utils.elementsAreUnique(new ArrayList<Object>() {
-            {
-                add(nullObject);
-                add(nullObject);
-            }
-        }));
-    }
-
-    @Test
-    public void elementsAreUnique_someIdenticalTags_returnsFalse() {
-        // To compare tags only
-        assertFalse(Utils.elementsAreUnique(new ArrayList<Tag>() {
-            {
-                add(tag1);
-                add(tag2);
-                add(tag2);
-            }
-        }));
-
-        assertFalse(Utils.elementsAreUnique(new ArrayList<Tag>() {
-            {
-                add(tag1);
-                add(tag1Copy);
-                add(tag2);
-            }
-        }));
-    }
-
-    @Test
-    public void elementsAreUnique_differentTags_returnsTrue() {
-        assertTrue(Utils.elementsAreUnique(new ArrayList<Tag>() {
-            {
-                add(tag1);
-                add(tag2);
-            }
-        }));
-
-        assertTrue(Utils.elementsAreUnique(new ArrayList<Tag>() {
-            {
-                add(tag1);
-                add(tag2);
-                add(null);
-            }
-        }));
+    void assertNotUnique(Object... objects) {
+        ArrayList<Object> list = new ArrayList<Object>(Arrays.asList(objects));
+        assertFalse(Utils.elementsAreUnique(list));
     }
 
 }
--- a/test/java/seedu/addressbook/common/UtilsTest.java
+++ b/test/java/seedu/addressbook/common/UtilsTest.java
@@ -3,21 +3,18 @@ package seedu.addressbook.common;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 
 import org.junit.Test;
-
-import seedu.addressbook.data.tag.Tag;
 
 public class UtilsTest {
     private static final Object NULL_OBJECT = null;
 
     @Test
     public void isAnyNull() {
-        ArrayList<String> emptyList = new ArrayList<String>();
-        ArrayList<String> nonEmptyList = new ArrayList<String>();
-        nonEmptyList.add("A string");
+        List<Object> emptyList = createList();
+        List<Object> nonEmptyList = createList("A string");
 
         // empty list
         assertFalse(Utils.isAnyNull());
@@ -27,8 +24,8 @@ public class UtilsTest {
         assertFalse(Utils.isAnyNull(emptyList));
 
         // non empty list with just one at the beginning
-        assertTrue(Utils.isAnyNull(NULL_OBJECT));
-        assertTrue(Utils.isAnyNull(NULL_OBJECT, "", new Object()));
+ //       assertTrue(Utils.isAnyNull(null));
+        assertTrue(Utils.isAnyNull(NULL_OBJECT, "", new Object(), null));
         assertTrue(Utils.isAnyNull(NULL_OBJECT, emptyList, nonEmptyList));
 
         // non empty list with nulls in the middle
@@ -42,35 +39,33 @@ public class UtilsTest {
 
     @Test
     public void elementsAreUnique() throws Exception {
-        Tag tag1 = new Tag("tag1"),
-            tag1Copy = new Tag("tag1"),
-            tag2 = new Tag("tag2");
-
         // empty list
         assertAreUnique();
 
         // only one object
         assertAreUnique(NULL_OBJECT);
-        assertAreUnique(tag1);
+        assertAreUnique(1);
 
         // all objects unique
         assertAreUnique("abc", "ab", "a");
-        assertAreUnique(tag1, tag2);
+        assertAreUnique(1, 2);
 
         // some identical objects
         assertNotUnique("abc", "", "abc", "ABC");
-        assertNotUnique(NULL_OBJECT, tag1, tag1Copy, tag2);
+        assertNotUnique(NULL_OBJECT, 1, new Integer(1));
         assertNotUnique("", "abc", "a", "abc");
         assertNotUnique(NULL_OBJECT, "a", "b", NULL_OBJECT);
     }
 
     private void assertAreUnique(Object... objects) {
-        ArrayList<Object> list = new ArrayList<Object>(Arrays.asList(objects));
-        assertTrue(Utils.elementsAreUnique(list));
+        assertTrue(Utils.elementsAreUnique(createList(objects)));
     }
 
     private void assertNotUnique(Object... objects) {
-        ArrayList<Object> list = new ArrayList<Object>(Arrays.asList(objects));
-        assertFalse(Utils.elementsAreUnique(list));
+        assertFalse(Utils.elementsAreUnique(createList(objects)));
+    }
+    
+    private List<Object> createList(Object... objects) {
+        return Arrays.asList(objects);
     }
 }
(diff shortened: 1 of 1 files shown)


test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,167 @@
...
+            return new Tag(tagName);
+        } catch (IllegalValueException e) {
+            throw new RuntimeException("test tag name should be valid", e);
+        }
+    }
+
+    @Test
+    public void isAnyNull_emptyInput_returnsFalse() {
+        assertFalse(Utils.isAnyNull());
+    }
```
damithc: `isAnyNull` is a simple method. Can be tested using one test method. Cases to test:
No nulls: 
* any non-empty list, 
* an empty list
With nulls: 
* non empty list with just one at the beginning (to check for bugs missing the first item in the list), 
* list with multiple objects, couple of nulls in the middle
* non empty list with one null as the last element (to check for bugs missing last item in the list)

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,167 @@
...
+        assertFalse(Utils.isAnyNull(emptyList, "A string"));
+    }
+
+    @Test
+    public void elementsAreUnique_emptyCollection_returnsTrue() {
+        assertTrue(Utils.elementsAreUnique(new ArrayList<>()));
+    }
+
+    @Test
+    public void elementsAreUnique_singleElement_returnsTrue() {
```
damithc: Can it be consolidated like this?
```java
@Test
public void elementsAreUnique() {
   //create test objects obj1, obj2, obj3, obj1copy
    assertAreUnique() //empty list
    assertAreUnique(obj1); //only one object
    assertAreUnique(obj1, obj2, obj3); // all objects unique
    assertNotUnique(obj1, obj2, obj1copy); //...
}

void assertAreUnique(Object... objects){
    ...
}
```

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: noted

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: noted

zzzzwen:
Updated according to reviews

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,89 @@
...
+        tag1 = buildTag("tag1");
+        tag1Copy = buildTag("tag1");
+        tag2 = buildTag("tag2");
+    }
+
+    private static Tag buildTag(String tagName) {
+        try {
+            return new Tag(tagName);
+        } catch (IllegalValueException e) {
+            throw new RuntimeException("test tag name should be valid", e);
```
damithc: You can use `fail()` here instead.

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,89 @@
...
+    private static Tag tag1;
+    private static Tag tag1Copy;
+    private static Tag tag2;
+
+    @Before
+    public void setup() {
+        tag1 = buildTag("tag1");
+        tag1Copy = buildTag("tag1");
+        tag2 = buildTag("tag2");
+    }
```
damithc: As we have only one test method using these, may be we should put them inside that method itself?

test/java/seedu/addressbook/common/UtilsTest.java
damithc: or just add a `throws Exception` to the method signature. If the exception happens, it will result in a test failure, which is what we want.

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Arrays;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import seedu.addressbook.data.exception.IllegalValueException;
+import seedu.addressbook.data.tag.Tag;
+
+public class UtilsTest {
+    private static Object nullObject = null;
```
damithc: declare as constant `NULL_OBJECT`?

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,89 @@
...
+        assertAreUnique(nullObject);
+        assertAreUnique(tag1);
+
+        //all objects unique
+        assertAreUnique("abc", "ab", "a");
+        assertAreUnique(tag1, tag2);
+
+        //some identical objects
+        assertNotUnique("abc", "", "abc");
+        assertNotUnique(tag1, tag1Copy, tag2);
```
damithc: In both cases one of the copies is the first object in the list. Chances of finding unexpected bugs can increase if you vary such conditions between test cases. e.g. What if there is a bug in the method that checks uniqueness against the first object only?

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: ok

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: Ya true should put inside the method which only uses them

zzzzwen:
some more changes

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,76 @@
...
+        // non empty list with one null as the last element
+        assertTrue(Utils.isAnyNull("", new Object(), NULL_OBJECT));
+        assertTrue(Utils.isAnyNull(nonEmptyList, emptyList, NULL_OBJECT));
+    }
+
+    @Test
+    public void elementsAreUnique() throws Exception {
+        Tag tag1 = new Tag("tag1"),
+            tag1Copy = new Tag("tag1"),
+            tag2 = new Tag("tag2");
```
damithc: Indentation problem

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        // some identical objects
+        assertNotUnique("abc", "", "abc", "ABC");
+        assertNotUnique(NULL_OBJECT, tag1, tag1Copy, tag2);
+        assertNotUnique("", "abc", "a", "abc");
+        assertNotUnique(NULL_OBJECT, "a", "b", NULL_OBJECT);
+    }
+
+    private void assertAreUnique(Object... objects) {
+        ArrayList<Object> list = new ArrayList<Object>(Arrays.asList(objects));
```
damithc: Duplicated in line 73. Extract to `createList(Object... objects)`? It can be used in lines 18 and 19 too.

damithc:
@pyokagan can have a look to see if I miss anything?

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,76 @@
...
+import seedu.addressbook.data.tag.Tag;
+
+public class UtilsTest {
+    private static final Object NULL_OBJECT = null;
+
+    @Test
+    public void isAnyNull() {
+        ArrayList<String> emptyList = new ArrayList<String>();
+        ArrayList<String> nonEmptyList = new ArrayList<String>();
+        nonEmptyList.add("A string");
```
pyokagan: Maybe we may be testing if `isAnyNull()` peeks inside containers to check if any of their containing elements are `null`. In that case, we should be adding `nulls` into these ArrayLists.

test/java/seedu/addressbook/common/UtilsTest.java
pyokagan: Why not just use `Arrays.asList()` directly? They are already `java.util.List`.

test/java/seedu/addressbook/common/UtilsTest.java
pyokagan: Should a `commons` test be depending on classes in the ~`model`~ `data` component?

Maybe we should stick with using more basic objects like strings.

If we want to test if `elementsAreUnique()` is calling the `#equals()` and `#hashcode()` methods, then we should use custom stubs for that.

pyokagan: What's the indentation problem?

damithc: Oh, I didn't notice it's intentional as this form of variable initialization is not common in java. I've seen people recommending one declaration per line e.g. Google's style guide https://google.github.io/styleguide/javaguide.html

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.junit.Test;
+
+import seedu.addressbook.data.tag.Tag;
+
+public class UtilsTest {
+    private static final Object NULL_OBJECT = null;
```
damithc: Yes, using `null` should be fine. If the meaning is clear, it's no longer a 'magic object'.

test/java/seedu/addressbook/common/UtilsTest.java
damithc: RE why test empty list, if the SUT is likely to have loops, it's good to throw in cases that can push the loop to its boundaries, such as 0 iterations. It's also confirming that given 0 elements the method returns false, which some developers can get wrong. Also, if I'm doing TDD, that would be my first test case (starting with the simplest case). Of course given the simple implementation of the method some of these cases seem redundant, but I prefer test cases to be written in black box fashion to future proof the tests.

test/java/seedu/addressbook/common/UtilsTest.java
damithc: Yes, it's best if this class doesn't depend on Model.

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: This is because the original `elementsAreUnique()` method takes in `Container<E>`, the `java.util.List` class does not implement such interface and thus cannot be used, so I used arraylist

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: is it necessary to check whether this method goes into the container to check? or just check whether container is not null

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: I'll remove the test for tags

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: ya true, I did  that because arraylist more intuitive for me, will change

test/java/seedu/addressbook/common/UtilsTest.java
pyokagan: @damithc Oh, I'm not disputing that :-). I'm just wondering why we are testing with `emptyList` and `nonEmptyList` **ArrayLists**, when `isAnyNull()` takes an `Object...`

pyokagan: > Why do we need to test "empty list", "non empty list"?

Yeah, this is badly worded. Sorry for the confusion.

pyokagan: @zzzzwen If you want to do the "isAnyNull() looks inside containers" check, then this line needs to be
```java
nonEmptyList.add(null);
```

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    public void isAnyNull() {
+        ArrayList<String> emptyList = new ArrayList<String>();
```
pyokagan: If we are doing the "isAnyNull() looks inside containers" thing, then this variable (`emptyList`) is not needed, since it does not tell us anything:

* If `isAnyNull()` looks inside containers, then `isAnyNull(emptyList)` will return false.
* If `isAnyNull()` does not look inside containers, then `isAnyNull(emptyList)` will still return false.

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,76 @@
...
+        ArrayList<String> nonEmptyList = new ArrayList<String>();
+        nonEmptyList.add("A string");
+
+        // empty list
+        assertFalse(Utils.isAnyNull());
+
+        // Any non-empty list
+        assertFalse(Utils.isAnyNull(new Object(), new Object()));
+        assertFalse(Utils.isAnyNull(emptyList));
```
pyokagan: If you want to do the "isAnyNull() looks inside containers" check, then we should use the `nonEmptyList` here.

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        // non empty list with just one at the beginning
+        assertTrue(Utils.isAnyNull(NULL_OBJECT));
+        assertTrue(Utils.isAnyNull(NULL_OBJECT, "", new Object()));
```
pyokagan: If you want to do tests with different kinds of objects, e.g. strings, containers etc., then you MUST ensure that these objects do not trigger `isAnyNull()`, otherwise we may get false negatives.

e.g. If you want to use `""` here, then in your `// Any non-empty list` tests above, you must add:
```java
assertFalse(Utils.isAnyNull(""));
```

test/java/seedu/addressbook/common/UtilsTest.java
pyokagan: Oh, it seems like this is another icky Java thing :-)

Yeah, if we want to call `isAnyNull(null)`, we need to make it explicit by using `isAnyNull((Object)null)`.

zzzzwen:
Some more changes

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: ok

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private void assertAreUnique(Object... objects) {
+        assertTrue(Utils.elementsAreUnique(createList(objects)));
+    }
+
+    private void assertNotUnique(Object... objects) {
+        assertFalse(Utils.elementsAreUnique(createList(objects)));
+    }
+
```
pyokagan: Trailing whitespace.

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,73 @@
...
+        assertTrue(Utils.isAnyNull(null, new Object(), new Object()));
+
+        // non empty list with nulls in the middle
+        assertTrue(Utils.isAnyNull(new Object(), null, null, "test"));
+        assertTrue(Utils.isAnyNull("", null, new Object()));
+
+        // non empty list with one null as the last element
+        assertTrue(Utils.isAnyNull("", new Object(), null));
+        assertTrue(Utils.isAnyNull(new Object(), new Object(), null));
+
```
pyokagan: Trailing whitespace.

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,73 @@
...
+        assertFalse(Utils.isAnyNull());
+
+        // Any non-empty list
+        assertFalse(Utils.isAnyNull(new Object(), new Object()));
+        assertFalse(Utils.isAnyNull("test"));
+        assertFalse(Utils.isAnyNull(""));
+
+        // non empty list with just one at the beginning
+        assertTrue(Utils.isAnyNull((Object)null));
+        assertTrue(Utils.isAnyNull(null, "", new Object(), null));
```
pyokagan: This group says `non empty list with just one at the beginning`, but there is a `null` at the end?

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,73 @@
...
+    private void assertAreUnique(Object... objects) {
+        assertTrue(Utils.elementsAreUnique(createList(objects)));
+    }
+
+    private void assertNotUnique(Object... objects) {
+        assertFalse(Utils.elementsAreUnique(createList(objects)));
+    }
+    
+    private List<Object> createList(Object... objects) {
+        return Arrays.asList(objects);
```
pyokagan: At this point might as well use `Arrays.asList()` directly.

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+        // all objects unique
+        assertAreUnique("abc", "ab", "a");
+        assertAreUnique(1, 2);
+
+        // some identical objects
+        assertNotUnique("abc", "", "abc", "ABC");
+        assertNotUnique(null, 1, new Integer(1));
+        assertNotUnique("", "abc", "a", "abc");
+        assertNotUnique(null, "a", "b", null);
```
pyokagan: To make the tests clearer, maybe you could e.g., for this test `assertNotUnique(null, "a", "b", null)`, add a test above with `assertNotUnique(null, null)`, to make it clear that you are testing that `elementsAreUnique()` still works even if there are unique elements mixed in.

Also, it aids as a check against false negatives (e.g., to check that `elementsAreUnique(null, "a", "b", null)` return false because `elementsAreUnique(null, null)` return false and not because `elementsAreUnique(null, "a")` returns false).

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: with one at beginning also includes null in the end? or should I split them?

zzzzwen: oh I think i should edit this and keep only one at the front

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: May feel a bit hard to read, I will change it first to see effect

test/java/seedu/addressbook/common/UtilsTest.java
zzzzwen: ok

zzzzwen:
Updated, but I still think that `Arrays.asList()` needs to be separated out

test/java/seedu/addressbook/common/UtilsTest.java
pyokagan: > oh I think i should edit this and keep only one at the front

Yes

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private void assertAreUnique(Object... objects) {
+        assertTrue(Utils.elementsAreUnique(Arrays.asList(objects)));
+    }
+
+    private void assertNotUnique(Object... objects) {
+        assertFalse(Utils.elementsAreUnique(Arrays.asList(objects)));
+    }
+}
```
pyokagan: Need a newline at end of file.

zzzzwen:
Updated and this should be fine

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,74 @@
...
+        // empty list
+        assertFalse(Utils.isAnyNull());
+
+        // Any non-empty list
+        assertFalse(Utils.isAnyNull(new Object(), new Object()));
+        assertFalse(Utils.isAnyNull("test"));
+        assertFalse(Utils.isAnyNull(""));
+
+        // non empty list with just one at the beginning
+        assertTrue(Utils.isAnyNull((Object)null));
```
pyokagan: Typecast needs to be followed by whitespace e.g. `(Object) null`

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertFalse(Utils.isAnyNull(nullList));
+    }
+
+    @Test
+    public void elementsAreUnique() throws Exception {
+        // empty list
+        assertAreUnique();
+
+        // only one object
+        assertAreUnique((Object)null);
```
pyokagan: Typecast needs to be followed by whitespace e.g. `(Object) null`

pyokagan:
@zzzzwen If you're free, maybe port these tests to Addressbook-Level4 as well :-)

zzzzwen:
changed, what do you mean by "port"? Is it a git feature? Or I just copy the test file and create another PR in level4

damithc:
@zzzzwen yes, just copy manually and create a new issue+PR

zzzzwen:
Done for level3 and 4
https://github.com/se-edu/addressbook-level3/pull/59
https://github.com/se-edu/addressbook-level4/pull/234

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,74 @@
...
+    public void isAnyNull() {
+        // empty list
+        assertFalse(Utils.isAnyNull());
+
+        // Any non-empty list
+        assertFalse(Utils.isAnyNull(new Object(), new Object()));
+        assertFalse(Utils.isAnyNull("test"));
+        assertFalse(Utils.isAnyNull(""));
+
+        // non empty list with just one at the beginning
```
damithc: `one null`?

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+        // non empty list with nulls in the middle
+        assertTrue(Utils.isAnyNull(new Object(), null, null, "test"));
+        assertTrue(Utils.isAnyNull("", null, new Object()));
+
+        // non empty list with one null as the last element
+        assertTrue(Utils.isAnyNull("", new Object(), null));
+        assertTrue(Utils.isAnyNull(new Object(), new Object(), null));
+
+        // test whether method peeks into the elements inside container
```
damithc: Better to be precise
`// confirms nulls inside the list are not considered`

test/java/seedu/addressbook/common/UtilsTest.java
```diff
@@ -0,0 +1,74 @@
...
+        // some identical objects
+        assertNotUnique("abc", "abc");
+        assertNotUnique("abc", "", "abc", "ABC");
+        assertNotUnique("", "abc", "a", "abc");
+
+        assertNotUnique(1, new Integer(1));
+        assertNotUnique(null, 1, new Integer(1));
+
+        assertNotUnique(null, null);
+        assertNotUnique(null, "a", "b", null);
```
damithc: It seems the rest of the method has one comment for each code block. Either add comments for these two blocks or merge them with the block above?

zzzzwen:
Updated comments

PR #128, opened 2016-12-22
Add unit tests for StorageFile class #104

### LINKED ISSUE #104: Add unit tests for StorageFile class
as per title

zzzzwen:
Fix #104 

Some format issues still.
I wonder if unit test on `StorageFile` can depend on `AddressBook` class or it is necessary to create another 'AddressBook' stub for this test though I only uses empty `AddressBook.`


zzzzwen:
Ready for review

test/java/seedu/addressbook/storage/InvalidData.txt
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<AddressBook>
+    <persons>
+        <name>John Doe</name>
+
+        <email isPrivate="false">johnd@gmail.com</email>
+        <address isPrivate="false">John street, block 123, #01-01</address>
+    </persons>
+</AddressBook>
```
damithc: Is this meant to be valid xml but invalid data, or invalid xml? Either way, make it more obvious? e.g use content that is obviously invalid. For example, this is clearly not even xml
```
This is some random text, not even xml
``` 
This is valid xml, but invalid data
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Invalid>
    <data>blah</data>
</Invalid>
```

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,95 @@
...
+        //the field in xml data does not match the AddressBook class, exception thrown 
+        StorageFile storage = new StorageFile(TEST_DATA_FOLDER + "/InvalidData.txt");
+        thrown.expect(StorageOperationException.class);
+        storage.load();
+    }
+
+    @Test
+    public void load_validFormat() throws StorageOperationException, IllegalValueException {
+        StorageFile storage = new StorageFile(TEST_DATA_FOLDER + "/ValidData.txt");
+        AddressBook ab = storage.load();
```
damithc: With a utility method, you can shorten this code like this:
```java
getStorage("/ValidData.txt").load();
```

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,95 @@
...
+        Person testPerson = new Person(new Name("John Doe"), 
+                                       new Phone("98765432", false),
+                                       new Email("johnd@gmail.com", false), 
+                                       new Address("John street, block 123, #01-01", false),
+                                       new UniqueTagList(Collections.emptySet()));
+        
+        // ensure loaded AddressBook Object is properly constructed with sample data
+        assertTrue(ab.getAllPersons().contains(testPerson));
```
damithc: Do a stronger comparison between the expected addressbook and the actual? Can be done as a separate issue (just put a `//TODO` here.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Test
+    public void initStorageFile_nullFilePath_exceptionThrown() throws InvalidStorageFilePathException {
+        //creating StroageFile with null filepath
+        thrown.expect(NullPointerException.class);
+        new StorageFile(null);
+    }
+
+    @Test
+    public void initStorageFile_invalidFilePath_exceptionThrown() throws InvalidStorageFilePathException {
+        //creating StorageFile with file not ending with .txt
```
zzzzwen: So I just leave as .txt first

test/java/seedu/addressbook/storage/InvalidData.txt
zzzzwen: In my thought, here i only check whether the xml format is valid. i.e. have relevant <Name><phone>... pairs according to the AddressBook class but in this test, I'm not testing whether data is valid, i.e. not testing the situation where <phone>abcdefg</phone>

zzzzwen: In the `StorageFile` class, it also only converts directly from data to the relevant Object as long as the xml structure is correct

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: What is "stronger comparison" ? Is it comparing char by char?

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: To separate as a utility method here is a bit redundant here, because only in this single test case such utility method is applicable and `save_validAddressBook()` this single test case also make use of StorageFile Object. Is it necessary to make it as 

StorageFile storage = getStorage("/temp.txt")
storage.save(ab)

zzzzwen: `StorageFIle storage = getStorage("/temp.txt")` and `StorageFile storage = new StorageFile(TEST_DATA_FOLDER + "/temp.txt");` looks the same to me but the first one creates another method

zzzzwen: Ahhh, looks like use a utility method is simpler and I have changed

zzzzwen:
@damithc 
Updated, what do you mean by "Stronger comparison" ? In what ways should the files be compared

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: Yes.

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: I meant the the entire address book object has exactly the expected data and nothing else (i.e. do an `equals` rather than `contains`). The current test can pass even if the address book loaded has more than one person objects.

test/java/seedu/addressbook/storage/InvalidData.txt
damithc: I see. Why is this xml file invalid? It's not immediately clear.

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: Oh ok, will do that way

zzzzwen:
Updated

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,99 @@
...
+        // ensure loaded AddressBook Object is properly constructed with sample data
+        assertTrue(ab.getAllPersons().equals(list));
+    }
+
+    @Test
+    public void save_nullAddressBook_exceptionThrown() throws InvalidStorageFilePathException, StorageOperationException {
+        //save null AddressBook object to StorageFile
+        StorageFile storage = getStorage("/temp.txt");
+        thrown.expect(NullPointerException.class);
+        storage.save(null);
```
damithc: Can line 69 and 71 be combined?

test/java/seedu/addressbook/storage/InvalidData.txt
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<AddressBook>
+    <persons>
+		<invalid>data</invalid>
```
damithc: This indentation is wrong. Is that intentional?

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    public void initStorageFile_invalidFilePath_exceptionThrown() throws InvalidStorageFilePathException {
+        //creating StorageFile with file not ending with .txt
+        thrown.expect(IllegalValueException.class);
+        new StorageFile(TEST_DATA_FOLDER + "/InvalidFileName");
+    }
+
+    @Test
+    public void load_invalidFormat_exceptionThrown() throws InvalidStorageFilePathException, StorageOperationException {
+        //the field in xml data does not match the AddressBook class, exception thrown
```
damithc: `// The file contains valid xml data, but does not match the AddressBook class`

test/java/seedu/addressbook/storage/InvalidData.txt
zzzzwen: Oh, it's two tabs, i will change it to 8 spaces for this line

zzzzwen:
Updated

test/java/seedu/addressbook/storage/StorageFileTest.java
pyokagan: Shouldn't all these comments be encoded within the method name instead?

e.g. `initStorageFile_noTxtExtension_throwsIllegalValueException()`

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private StorageFile getStorage(String filename) throws InvalidStorageFilePathException {
+        return new StorageFile(TEST_DATA_FOLDER + filename);
+    }
+
+    private boolean compareFile(String file1, String file2) throws IOException {
+        String s1 = new String(Files.readAllBytes(Paths.get(TEST_DATA_FOLDER + file1)));
+        String s2 = new String(Files.readAllBytes(Paths.get(TEST_DATA_FOLDER + file2)));
+        return s1.equals(s2);
```
pyokagan: May wish to use `assertEquals()` directly here, JUnit will provide more useful error information such as a diff of how the two strings differ.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,98 @@
...
+        // ensure xml data for sample data is properly structured and saved
+        assertTrue(compareFile("/temp.txt", "/ValidData.txt"));
+    }
+
+    private StorageFile getStorage(String filename) throws InvalidStorageFilePathException {
+        return new StorageFile(TEST_DATA_FOLDER + filename);
+    }
+
+    private boolean compareFile(String file1, String file2) throws IOException {
+        String s1 = new String(Files.readAllBytes(Paths.get(TEST_DATA_FOLDER + file1)));
```
pyokagan: Use `Paths.get(TEST_DATA_FOLDER, file1)` rather than string concatenation?

Then you don't need to prefix all your file paths with `/`.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,98 @@
...
+                                       new UniqueTagList(Collections.emptySet()));
+        ab.addPerson(testPerson);
+        storage.save(ab);
+
+        // ensure xml data for sample data is properly structured and saved
+        assertTrue(compareFile("/temp.txt", "/ValidData.txt"));
+    }
+
+    private StorageFile getStorage(String filename) throws InvalidStorageFilePathException {
+        return new StorageFile(TEST_DATA_FOLDER + filename);
```
pyokagan: Use `Paths.get()` rather than string concatenation here as well? Or use `File.separator`.

Then you don't need to prefix all your file paths with `/`.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,98 @@
...
+import seedu.addressbook.data.person.Name;
+import seedu.addressbook.data.person.Person;
+import seedu.addressbook.data.person.Phone;
+import seedu.addressbook.data.person.UniquePersonList;
+import seedu.addressbook.data.tag.UniqueTagList;
+import seedu.addressbook.storage.StorageFile.InvalidStorageFilePathException;
+import seedu.addressbook.storage.StorageFile.StorageOperationException;
+
+public class StorageFileTest {
+    private static final String TEST_DATA_FOLDER = "./test/java/seedu/addressbook/storage";
```
pyokagan: Why the `./` in front? Shouldn't be needed.

(Yes, I know the other tests have this issue as well).

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,98 @@
...
+                                       new Address("John street, block 123, #01-01", false),
+                                       new UniqueTagList(Collections.emptySet()));
+        UniquePersonList list = new UniquePersonList(testPerson);
+
+        // ensure loaded AddressBook Object is properly constructed with sample data
+        assertTrue(ab.getAllPersons().equals(list));
+    }
+
+    @Test
+    public void save_nullAddressBook_exceptionThrown() throws InvalidStorageFilePathException, StorageOperationException {
```
pyokagan: Maybe we can just use `throws Exception` for all tests rather than listing all of them out?

@damithc thoughts?

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,98 @@
...
+        //save null AddressBook object to StorageFile
+        thrown.expect(NullPointerException.class);
+        getStorage("/temp.txt").save(null);
```
pyokagan: @damithc I think the previous version was better. e.g. `getStorage()` may be the one throwing the NPE.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,98 @@
...
+import seedu.addressbook.storage.StorageFile.StorageOperationException;
+
+public class StorageFileTest {
+    private static final String TEST_DATA_FOLDER = "./test/java/seedu/addressbook/storage";
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void initStorageFile_nullFilePath_exceptionThrown() throws InvalidStorageFilePathException {
```
pyokagan: @damithc I thought the conventional naming for tests of constructors is just `constructor`?

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: ok

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: I agree, previously only the `save` part is tested, the init of `StorageFile` is before `thrown.expect(NullPointerException.class); but I feel writing in this way is clearer?`

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: Yes, it should be `constructor`

zzzzwen:
Since the `StorageFile` is the only class file in `storage` package, so I think it is ok for this level 2 only to put  the test data within the same folder?

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: Yes, I agree. let's go back to the previous one.

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: `StorageFile` constructor only takes in filepath in `String` format, `Paths.get()` returns a `Path` Object which needs later to be converted to `String`, a bit redundant I think

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: Yes, that's what we do in other tests too. Using `throws Exception` results in less code and we don't really need to know exact types of exceptions thrown from test methods.

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: ok

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: changed, used `./` to notice myself it's a relative path

zzzzwen: but they are just the same

zzzzwen:
I have changed a bit of the path issues

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: So only for this case or remove the use of `getStorage `for all?

damithc: All cases where the same reasoning applies.

zzzzwen: So is it OK that I keep the `getStorage()` method but put it before `thrown.expect(NullPointerException.class)`, in this way less changes needed.

zzzzwen:
Updated

test/java/seedu/addressbook/storage/StorageFileTest.java
pyokagan: Not really. Putting `./` in front is putting a redundant path component. This is akin to constructing the string using `"test/" + "java/" + "seedu/" + "addressbook/" + "storage"` -- we still get the same string but the code is not succinct.

test/java/seedu/addressbook/storage/StorageFileTest.java
pyokagan: OK then.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+                                       new UniqueTagList(Collections.emptySet()));
+        UniquePersonList list = new UniquePersonList(testPerson);
+
+        // ensure loaded AddressBook Object is properly constructed with sample data
+        assertTrue(ab.getAllPersons().equals(list));
+    }
+
+    @Test
+    public void save_nullAddressBook_exceptionThrown() throws Exception {
+        // save null AddressBook object to StorageFile
```
pyokagan: Can remove this comment.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private StorageFile getStorage(String filename) throws Exception {
+        return new StorageFile(TEST_DATA_FOLDER + filename);
+    }
+
+    private void assertSaveSuccess(String file1, String file2) throws Exception {
+        String s1 = new String(Files.readAllBytes(Paths.get(TEST_DATA_FOLDER, file1)));
+        String s2 = new String(Files.readAllBytes(Paths.get(TEST_DATA_FOLDER, file2)));
+        assertTrue(s1.equals(s2));
```
pyokagan: Still not using `assertEquals()`?

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: ok

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: ah, I misread this line sorry

zzzzwen:
Updated, but I met a problem when calling assertEquals here

org.junit.ComparisonFailure: expected:<..." standalone="yes"?>[
<AddressBook>
    <persons>
        <name>John Doe</name>
        <phone isPrivate="false">98765432</phone>
        <email isPrivate="false">johnd@gmail.com</email>
        <address isPrivate="false">John street, block 123, #01-01</address>
    </persons>
</AddressBook>]
> but was:<..." standalone="yes"?>[
<AddressBook>
    <persons>
        <name>John Doe</name>
        <phone isPrivate="false">98765432</phone>
        <email isPrivate="false">johnd@gmail.com</email>
        <address isPrivate="false">John street, block 123, #01-01</address>
    </persons>
</AddressBook>
]
>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at seedu.addressbook.storage.StorageFileTest.assertSaveSuccess(StorageFileTest.java:95)
	at seedu.addressbook.storage.StorageFileTest.save_validAddressBook(StorageFileTest.java:85)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.rules.ExpectedException$ExpectedExceptionStatement.evaluate(ExpectedException.java:239)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:86)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:459)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:675)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:192)


It seems like the square brackets are not in correct position, when I run the actual program and paste the data inside, it works, but when I commit the change, the github doesn't show the change, why is this so? What are the square brackets used for in the xml file?

zzzzwen:
except the above `]` which is not in txt data file, the rest is done

damithc:
@zzzzwen for future reference, your branch name does not follow the naming convention.

test/java/seedu/addressbook/storage/InvalidData.txt
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<AddressBook>
+    <persons>
+        <invalid>data</invalid>
+    </persons>
+</AddressBook>
```
damithc: Test data should be in a different folder, not in the same folder as java files. See higher level addressbook projects for reference.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private StorageFile getStorage(String filename) throws Exception {
+        return new StorageFile(TEST_DATA_FOLDER + filename);
+    }
+
+    private void assertSaveSuccess(String file1, String file2) throws Exception {
+        String s1 = new String(Files.readAllBytes(Paths.get(TEST_DATA_FOLDER, file1)));
+        String s2 = new String(Files.readAllBytes(Paths.get(TEST_DATA_FOLDER, file2)));
+        assertEquals(s1, s2);
```
damithc: The assertion failure is caused by line separators (different OS's use different line endings).  The square brackets you see can be ignored (they are inserted by Eclipse to mark the parts that differ).

zzzzwen:
@damithc oh, I didn't notice the naming convention for branch name, so should this branch be called "104-unit-test-StorageFile" ?

zzzzwen:
Test files moved to folders following level 4 format

damithc:
>so should this branch be called "104-unit-test-StorageFile" ?

Yes, something like that. Can leave it as it is for this PR. Remember to follow it for future PRs.

zzzzwen:
ok, any more changes?

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: You have to fix this problem.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+                                       new Email("johnd@gmail.com", false), 
+                                       new Address("John street, block 123, #01-01", false),
+                                       new UniqueTagList(Collections.emptySet()));
+        ab.addPerson(testPerson);
+        storage.save(ab);
+
+        // ensure xml data for sample data is properly structured and saved
+        assertSaveSuccess("temp.txt", "ValidData.txt");
+    }
+
```
damithc: Put a comment to explain why `getPath` method is not tested.

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: If you can't find a way around this problem, I don't mind if you use the load method to verify the save method. Although not very desirable (a bug in load method can break tests on save method), it gets the job done and even can uncover unexpected bugs in the load method.
@pyokagan any input?

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void load_validFormat() throws Exception {
+        AddressBook ab = getStorage("/ValidData.txt").load();
+        Person testPerson = new Person(new Name("John Doe"), 
+                                       new Phone("98765432", false),
+                                       new Email("johnd@gmail.com", false), 
+                                       new Address("John street, block 123, #01-01", false),
+                                       new UniqueTagList(Collections.emptySet()));
+        UniquePersonList list = new UniquePersonList(testPerson);
+
+        // ensure loaded AddressBook Object is properly constructed with sample data
```
damithc: `sample data` --> `given data` or `test data`

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Test
+    public void load_validFormat() throws Exception {
+        AddressBook ab = getStorage("/ValidData.txt").load();
+        Person testPerson = new Person(new Name("John Doe"), 
+                                       new Phone("98765432", false),
+                                       new Email("johnd@gmail.com", false), 
+                                       new Address("John street, block 123, #01-01", false),
+                                       new UniqueTagList(Collections.emptySet()));
+        UniquePersonList list = new UniquePersonList(testPerson);
```
damithc: List having one object is a special case. You may want to have at least two objects to ensure that the method saves all objects, not just the first object.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+import seedu.addressbook.storage.StorageFile.StorageOperationException;
+
+public class StorageFileTest {
+    private static final String TEST_DATA_FOLDER = "test/data/StorageFileTest";
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void initStorageFile_nullFilePath_exceptionThrown() throws Exception {
```
damithc: Should be `constructor_...`

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void save_nullAddressBook_exceptionThrown() throws Exception {
+        StorageFile storage = getStorage("/temp.txt");
+        thrown.expect(NullPointerException.class);
+        storage.save(null);
+    }
+
+    @Test
+    public void save_validAddressBook() throws Exception {
+        StorageFile storage = getStorage("/temp.txt");
+        AddressBook ab = new AddressBook();
```
damithc: This line can be pushed down. No need to declare a variable until you need it.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Person testPerson = new Person(new Name("John Doe"), 
+                                       new Phone("98765432", false),
+                                       new Email("johnd@gmail.com", false), 
+                                       new Address("John street, block 123, #01-01", false),
+                                       new UniqueTagList(Collections.emptySet()));
```
damithc: This code is duplicated.

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: @damithc  Can I do a line by line comparison? In this way line separators may be removed

zzzzwen:
```
Scanner scan1 = new Scanner(Paths.get(TEST_DATA_FOLDER, file1));
Scanner scan2 = new Scanner(Paths.get(TEST_DATA_FOLDER, file2));
while (scan1.hasNext() && scan2.hasNext()) {
    assertEquals(scan1.nextLine(), scan2.nextLine());
}
```

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: `getPath` method here refers to the `Paths.get()` method in` assertSaveSuccess()` right

damithc: No, there is a [public getPath method in StorageFile class](https://github.com/se-edu/addressbook-level2/blob/master/src/seedu/addressbook/storage/StorageFile.java#L152-L154) that needs to be covered (a test class is supposed to test all public methods of the class being tested). It's a trivial method so there is no need to test it. But we can record the reason here for completeness' sake.

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: Yes, that's one way. In any case, this method can be moved to TestUtil class.
BTW, your code sample will still pass even if 2nd file has additional lines at the end.

zzzzwen: ok, I did the checking for extra lines in actual code

zzzzwen:
Updated

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -60,4 +65,17 @@ public static Person generateTestPerson() {
             return null;
         }
     }
+    
+    public static void compareFiles(Path path1, Path path2) throws IOException {
```
damithc: `isSameTextIgnoreLineEndings(): boolean`? Just return a boolean instead of coupling this file to JUnit.
Add a header comment.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private void assertSaveSuccess(String file1, String file2) throws Exception {
+        TestUtil.compareFiles(Paths.get(TEST_DATA_FOLDER, file1), Paths.get(TEST_DATA_FOLDER, file2));
+    }
+
+    private StorageFile getStorage(String filename) throws Exception {
+        return new StorageFile(TEST_DATA_FOLDER + filename);
+    }
+
+    private Person[] getTestPerson() throws Exception {
```
damithc: Shouldn't this be `getTestPersons()`?

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,102 @@
...
+        thrown.expect(NullPointerException.class);
+        storage.save(null);
+    }
+
+    @Test
+    public void save_validAddressBook() throws Exception {
+        AddressBook ab = new AddressBook();
+        Person[] testPerson = getTestPerson();
+        ab.addPerson(testPerson[0]);
+        ab.addPerson(testPerson[1]);
```
damithc: Why can't the `getTestPerson()` return an AddressBook? That way, these 4 lines can be avoided altogether.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,102 @@
...
+        storage.load();
+    }
+
+    @Test
+    public void load_validFormat() throws Exception {
+        AddressBook ab = getStorage("/ValidData.txt").load();
+        UniquePersonList list = new UniquePersonList(getTestPerson());
+
+        // ensure loaded AddressBook Object is properly constructed with test data
+        assertTrue(ab.getAllPersons().equals(list));
```
damithc: 1. `assertEquals(list, getTestPerson())`
2. Why not compare the two `AddressBook` objects themselves?

test/java/seedu/addressbook/util/TestUtil.java
pyokagan: > isSameTextIgnoreLineEndings(): boolean

Or just `compareTextFiles()`, which uses the "text" mode terminology. Common File IO APIs (e.g. C's fopen() or python's open()) allow operating in "text" mode, where the file's bytes are decoded to the programming language's  native string type and line endings are normalized.

> Just return a boolean

I prefer using JUnit's assert methods so that JUnit can give us a clue on how the files differ.

If we are using JUnit's assert methods though, then this method should be called `assertTextFilesEqual()` or something.

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -60,4 +65,17 @@ public static Person generateTestPerson() {
...
+        Scanner scan1 = new Scanner(path1);
+        Scanner scan2 = new Scanner(path2);
```
pyokagan: Seems like [`Files.readAllLines()`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#readAllLines-java.nio.file.Path-java.nio.charset.Charset-) will be much more succinct.

Take note of the decoding charset though. I think it needs to be `Charset.defaultCharset()` rather than UTF-8.

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: @damithc `equals` method is not overwritten in `AddressBook` class,  I try to minimise the modification of other classes when writing tests and so I'm comparing person lists

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: oh ya

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: or if want to just return `AddressBook` I will copy the `equals` method from level 4

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: I did not compare the `AddressBook` object and mentioned above, so here it also needs same person data, I did not return as `AddressBook` object but as list

test/java/seedu/addressbook/util/TestUtil.java
damithc: 1. That's a valid point about using JUnit assert.  
2. I'm ok with using `Text` in the name to indicate text mode. `assertTextFilesEqual()` sounds good.

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: Using equals can be done separately. Just add a `//TODO:` comment.

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: In the other method, you can retrieve the person list from `AddressBook` object, right?

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: if in this way this test will likely to fail before equals method been added, is this fine?

zzzzwen: oh nvm, I will keep as comparing person list

zzzzwen:
Updated

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -60,4 +65,13 @@ public static Person generateTestPerson() {
             return null;
         }
     }
+
+    /**
+     * Assert whether the data in the two given files are the same.
```
damithc: `data` --> `text`
It's better to mention that any difference in line endings are ignored.

damithc:
@pyokagan can have a look?

test/java/seedu/addressbook/util/TestUtil.java
zzzzwen: ok

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,99 @@
...
+        getStorage("/temp.txt").save(ab);
+
+        // ensure xml data for sample data is properly structured and saved
+        assertSaveSuccess("temp.txt", "ValidData.txt");
+    }
+
+    // getPath() method in StorageFile class is trivial so it is not tested
+
+    private void assertSaveSuccess(String file1, String file2) throws Exception {
+        TestUtil.assertTextFilesEqual(Paths.get(TEST_DATA_FOLDER, file1), Paths.get(TEST_DATA_FOLDER, file2));
```
pyokagan: I wonder if we should `import static TestUtil.assertTextFilesEqual;` to be consistent with how we `import static org.junit.Assert.assertEquals;`

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,99 @@
...
+        storage.load();
+    }
+
+    @Test
+    public void load_validFormat() throws Exception {
+        AddressBook actualAB = getStorage("/ValidData.txt").load();
+        AddressBook expectedAB = getTestAddressBook();
+
+        // ensure loaded AddressBook Object is properly constructed with test data
+        //TODO overwrite equals method in AddressBook class and replace with equals method below
```
pyokagan: `// TODO: overwrite...` (space after `//` and `:` after `TODO`)

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,99 @@
...
+        thrown.expect(StorageOperationException.class);
+        storage.load();
+    }
+
+    @Test
+    public void load_validFormat() throws Exception {
+        AddressBook actualAB = getStorage("/ValidData.txt").load();
+        AddressBook expectedAB = getTestAddressBook();
+
+        // ensure loaded AddressBook Object is properly constructed with test data
```
pyokagan: Remove `Object`

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,99 @@
...
+        StorageFile storage = getStorage("/temp.txt");
+        thrown.expect(NullPointerException.class);
+        storage.save(null);
+    }
+
+    @Test
+    public void save_validAddressBook() throws Exception {
+        AddressBook ab = getTestAddressBook();
+
+        getStorage("/temp.txt").save(ab);
```
pyokagan: Very important: shouldn't we be saving this in either a temporary directory or a sandbox directory that is gitignored?

zzzzwen: @pyokagan yes, I think should be. Or can I just add the filepath of the temp data to .gitignore file?

zzzzwen:
Updated

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -60,4 +65,13 @@ public static Person generateTestPerson() {
             return null;
         }
     }
+
+    /**
+     * Assert whether the text in the two given files are the same.
```
damithc: `Asserts`

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,111 @@
...
+        return new StorageFile(TEST_DATA_FOLDER + fileName);
+    }
+
+    private StorageFile getTempStorage() throws Exception {
+        return new StorageFile(getTempFilePath("temp.txt"));
+    }
+
+    private String getTempFilePath(String fileName) {
+        return testFolder.getRoot().getPath() + fileName;
+    }
```
damithc: Can inline this method? Doesn't add much value.

zzzzwen:
Updated

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -79,19 +79,16 @@ public void save_validAddressBook() throws Exception {
...
         return new StorageFile(TEST_DATA_FOLDER + fileName);
     }
 
     private StorageFile getTempStorage() throws Exception {
-        return new StorageFile(getTempFilePath("temp.txt"));
-    }
-
-    private String getTempFilePath(String fileName) {
-        return testFolder.getRoot().getPath() + fileName;
+        return new StorageFile(testFolder.getRoot().getPath() + "temp.txt");
```
damithc: You have `temp.txt` in two places. Define as a constant?

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    // getPath() method in StorageFile class is trivial so it is not tested
+
+    private void assertSaveSuccess(String fileName) throws Exception {
+        assertTextFilesEqual(Paths.get(testFolder.getRoot().getPath() + "temp.txt"), 
+                             Paths.get(TEST_DATA_FOLDER, fileName));
+    }
+
+    private StorageFile getStorage(String fileName) throws Exception {
+        return new StorageFile(TEST_DATA_FOLDER + fileName);
```
pyokagan: You can just `TEST_DATA_FOLDER + "/" + fileName` and then you won't have to have all the slashes in your file names in your test code.

zzzzwen:
Updated, is it always necessary to change repeated String/Objects as constant?

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+        // ensure xml data for sample data is properly structured and saved
+        assertSaveSuccess("ValidData.txt");
+    }
+
+    // getPath() method in StorageFile class is trivial so it is not tested
+
+    private void assertSaveSuccess(String fileName) throws Exception {
+        assertTextFilesEqual(Paths.get(testFolder.getRoot().getPath() + TEST_TEMP), 
+                             Paths.get(TEST_DATA_FOLDER, fileName));
```
pyokagan: Not DRY: The code for constructing the temporary storage path (`testFolder.getRoot().getPath() + TEST_TEMP`) is repeated in `getTempStorage()`.

The code for constructing the test data storage path (`Paths.get(TEST_DATA_FOLDER, fileName)`) is repeated in `getStorage()`.

pyokagan: Also, the string representation of `testFolder.getRoot().getPath()` does not have a trailing slash, so I get file paths like `/tmp/junit7267453095673558989temp.txt` on my system and JUnit does not clean up the temporary files properly.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,109 @@
...
+        storage.save(null);
+    }
+
+    @Test
+    public void save_validAddressBook() throws Exception {
+        AddressBook ab = getTestAddressBook();
+
+        getTempStorage().save(ab);
+
+        // ensure xml data for sample data is properly structured and saved
```
pyokagan: This should be in the javadoc of `assertSaveSuccess()` rather than here.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,109 @@
...
+        assertSaveSuccess("ValidData.txt");
```
pyokagan: Method name doesn't seem to describe what we want to test for very well.

All we want to do is to compare the contents of two files, it is just that we can't easily get the paths of the files we want to compare (since we don't have suitable methods for them).

One way out of this is to:

Introduce a method `assertStorageFilesEqual(StorageFile, StorageFile)`:
```java
    private void assertStorageFilesEqual(StorageFile a, StorageFile b) throws Exception {
        assertTextFilesEqual(Paths.get(a.getPath()), Paths.get(b.getPath()));
    }
```

We can then replace the above call to `assertSaveSuccess()` with:
```java
assertStorageFilesEqual(tempStorage, getStorage("ValidData.txt"));
```

Which is much less opaque.

zzzzwen: @pyokagan ah, I forget about the `getPath()` method in `StorageFile` since it is trivival, writing in this way looks cleaner and easier to understand. Thanks

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: @pyokagan if with slash, i.e. `/tmp/junit7267453095673558989/temp.txt` the Junit will clean up the temp files properly is it?

zzzzwen:
Updated

test/java/seedu/addressbook/storage/StorageFileTest.java
pyokagan: Yes, because all it does is to delete the `/tmp/junit7267453095673558989/` directory after test execution.

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,109 @@
...
+    public void save_nullAddressBook_exceptionThrown() throws Exception {
+        StorageFile storage = getTempStorage();
+        thrown.expect(NullPointerException.class);
+        storage.save(null);
+    }
+
+    @Test
+    public void save_validAddressBook() throws Exception {
+        AddressBook ab = getTestAddressBook();
+        getTempStorage().save(ab);
```
pyokagan: The following looks easier to read for me but this is just personal taste:
```java
StorageFile storage = getTempStorage();
storage.save(ab);
assertStorageFilesEqual(storage, ...);
```

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -60,4 +65,13 @@ public static Person generateTestPerson() {
...
         }
     }
+
+    /**
+     * Asserts whether the text in the two given files are the same.
+     */
+    public static void assertTextFilesEqual(Path path1, Path path2) throws IOException {
+        List<String> list1 = Files.readAllLines(path1, Charset.defaultCharset());
+        List<String> list2 = Files.readAllLines(path2, Charset.defaultCharset());
+        assertEquals(list1, list2);
```
pyokagan: Maybe make this
```java
assertEquals(String.join("\n", list1), String.join("\n", list2));
```
so that the assertion error message is more meaningful.

(If we are comparing lists, JUnit prints the full contents of the lists. If we are comparing strings, JUnit only prints the differences).

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,109 @@
...
+import seedu.addressbook.data.person.Person;
+import seedu.addressbook.data.person.Phone;
+import seedu.addressbook.data.tag.Tag;
+import seedu.addressbook.data.tag.UniqueTagList;
+import seedu.addressbook.storage.StorageFile.StorageOperationException;
+import static seedu.addressbook.util.TestUtil.assertTextFilesEqual;
+
+public class StorageFileTest {
+    private static final String TEST_DATA_FOLDER = "test/data/StorageFileTest";
+    private static final String TEST_TEMP = "temp.txt";
```
pyokagan: I actually don't think this should be a constant as it's not very meaningful. It is just a temporary file name, it can be anything and readers of the code should not need to care (as opposed to `TEST_DATA_FOLDER` where it is important that readers of the code know its location since that is where the test data is stored).

zzzzwen: This constant is previously requested by Prof, I'm not sure whether it is better to inline or leave as constant

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: but `storage` variable is only used once so I think it is ok to be like this?

test/java/seedu/addressbook/util/TestUtil.java
zzzzwen: I'm not sure whether this is more useful or not, but I think that to have a full content is clearer to see because if you only show the differences, you may not know which `Person` the differences belong to, i.e. if only `Email` differs, the other information is not shown right?

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: Em, maybe needs a local variable to store it. Logically, to compare the `StorageFile` should compare the `StorageFile` object for temp and the valid data, but the `StorageFile` object for saving AddressBook and calling `assertStorageFilesEqual` differs since they are initialised by `getTempStorage()` twice. How do you think?

test/java/seedu/addressbook/util/TestUtil.java
pyokagan: > but I think that to have a full content is clearer to see

Usually it's not. Since this is a utility method, it is better to cater to the general case.

But really, you can try and see for yourself and report back the results :-)

> if you only show the differences, you may not know which Person the differences belong to, i.e. if only Email differs, the other information is not shown right?

JUnit shows some context with the differences.

zzzzwen: > org.junit.ComparisonFailure: expected:<...rivate="false">johnd[@gmail.com</email>
>         <address isPrivate="false">John street, block 123, #01-01</address>
>     </persons>
>     <persons>
>         <name>Betsy Crowe</name>
>         <phone isPrivate="true">1234567</phone>
>         <email isPrivate="false">betsycrowe@gmail.com</email>
>         <address isPrivate="true">Newgate Prison</address>
>         <tagged>friend</tagged>
>         <tagged>crim]inal</tagged>
>     </...> but was:<...rivate="false">johnd[f@gmail.com</email>
>         <address isPrivate="false">John street, block 123, #01-01</address>
>     </persons>
>     <persons>
>         <name>Betsy Crowe</name>
>         <phone isPrivate="true">123456227</phone>
>         <email isPrivate="false">betsycrowe@gmail.com</email>
>         <address isPrivate="true">Newgate Prison</address>
>         <tagged>friend</tagged>
>         <tagged>crimf]inal</tagged>
>     </...>

This is what JUnit gives, the first `Person` I only edited `Email `to wrong data but second one I edited both `Phone `and `Tag `to be wrong data. It is clear that first `Person` only gives one extra line but second gives a lot more. 

Since if it is general practice to compare string and only show differences, I will change to cater to general case.

test/java/seedu/addressbook/storage/StorageFileTest.java
pyokagan: My reasoning is that the current code requires the reader to put in *more effort* to understand how the test works. The reader needs to further understand that `getTempStorage()` returns new `StorageFile` objects, but they are of the same path, and that `StorageFile` will write to the same file, so it is therefore safe to use two different "temporary storage" objects in this test.

So, I prefer writing test code that is more obvious, which is why I recommended the code above.

test/java/seedu/addressbook/storage/StorageFileTest.java
damithc: I requested it because it was duplicated in two places and was not DRY. Changing one required changing the other.  If we keep the constant, the name should be improved. But I'm ok to inline it too.

zzzzwen: Previously it is used in two places, current version only uses this constant once, so I just inline it.

test/java/seedu/addressbook/storage/StorageFileTest.java
zzzzwen: ok

zzzzwen:
Updated

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,109 @@
...
+        StorageFile storage = getTempStorage();
+        storage.save(ab);
+
+        assertStorageFilesEqual(storage, getStorage("ValidData.txt"));
+    }
+
+    // getPath() method in StorageFile class is trivial so it is not tested
+
+    /**
+     * Ensure text file for test data is properly structured and saved.
```
damithc: `Ensures`

zzzzwen:
Updated

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,109 @@
...
+        StorageFile storage = getTempStorage();
+        storage.save(ab);
+
+        assertStorageFilesEqual(storage, getStorage("ValidData.txt"));
+    }
+
+    // getPath() method in StorageFile class is trivial so it is not tested
+
+    /**
+     * Ensures text file for test data is properly structured and saved.
```
pyokagan: Maybe remove this javadoc completely or describe what the method does (e.g. `Asserts that the contents of two storage files are the same.`)

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -60,4 +65,13 @@ public static Person generateTestPerson() {
             return null;
         }
     }
+
+    /**
+     * Asserts whether the text in the two given files are the same.
```
damithc: Add `Ignores any differences in line endings`?

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -0,0 +1,110 @@
...
+        storage.save(ab);
+
+        assertStorageFilesEqual(storage, getStorage("ValidData.txt"));
+    }
+
+    // getPath() method in StorageFile class is trivial so it is not tested
+
+    /**
+     * Asserts that the contents of two storage files are the same. Ensures text 
+     * file for test data is properly structured and saved.
```
damithc: I'm not sure if the second sentence is relevant. The header comment should not include information about the usage context.

zzzzwen: em, maybe delete it

zzzzwen:
Updated

pyokagan:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/se-edu/addressbook-level2/128)*

PR #131, opened 2016-12-29
Add unit tests for ViewAllCommand and ViewCommand classes #111

### LINKED ISSUE #111: Add unit tests for ViewAllCommand and ViewCommand classes
as per title

zzzzwen:
code ideas and format follow `DeleteCommandTest`

Adding header comments later

For ViewAll command, should I do it in a separate file or together with view command

Fixes #111


damithc:
@zzzzwen 

> For ViewAll command, should I do it in a separate file or together with view command

Usually, we have one test class for each functional class

zzzzwen:
Ready for review,

two tests are roughly the same

Here are some of my questions:

in `ViewCommandTest`:
`assertViewSuccess`, `assertViewErrorInvalidIndex`, `assertViewErrorPersonNotInAddressBook` are actually the same except the expected message is different, should I inline all three?

`assertCommandResult` this method can be quite general by just asserting the result message, if modify a bit, can move it to `TestUtil` class?

Same thing apply to `ViewAllCommandTest`

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,150 @@
...
+        Person johnDoe = new Person(new Name("John Doe"),
+                                    new Phone("98765432", false),
+                                    new Email("johnd@gmail.com", false),
+                                    new Address("John street, block 123, #01-01", false),
+                                    new UniqueTagList(Collections.emptySet()));
+        Person betsyCrowe = new Person(new Name("Betsy Crowe"),
+                                       new Phone("1234567", true),
+                                       new Email("betsycrowe@gmail.com", false),
+                                       new Address("Newgate Prison", true),
+                                       new UniqueTagList(new Tag("friend"), new Tag("criminal")));
```
damithc: Two persons is a bit a special case because the list doesn't have persons in the middle. A 3-person list is more general.

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+        emptyAddressBook = TestUtil.createAddressBook();
+        addressBook = TestUtil.createAddressBook(johnDoe, betsyCrowe);
+
+        emptyDisplayList = TestUtil.createList();
+        listWithAll = TestUtil.createList(johnDoe, betsyCrowe);
+        listWithSome = TestUtil.createList(betsyCrowe);
+    }
```
damithc: Extract into a `TypicalPersons` utility class?

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Test
+    public void viewAllCommand_invalidIndex_returnsInvalidIndexMessage() {
+        // empty addressbook
+        assertViewAllErrorInvalidIndex(emptyAddressBook, emptyDisplayList, 1);
+
+        // non-empty addressbook
+        assertViewAllErrorInvalidIndex(addressBook, listWithAll, -1);
+        assertViewAllErrorInvalidIndex(addressBook, listWithAll, 0);
+        assertViewAllErrorInvalidIndex(addressBook, listWithAll, 3);
```
damithc: if you use `size()+1` of the list instead of magic number `3`, the test will be less brittle (less likely to break in future).

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+        Command command = generateViewAllCommand(addressBook, list, index);
+        assertCommandResult(command, expectedMessage, addressBook);
+    }
+
+    /**
+     * Executes the command, and asserts the result message is as expected.
+     */
+    private void assertCommandResult(Command command, String expectedMessage,
+                                        AddressBook addressBook) {
```
damithc: Do you need this line break?

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+        return command;
+    }
+
+    /**
+     * Asserts that the details person at specific index can be successfully retrieved
+     * and displayed.
+     */
+    private void assertViewAllSuccess(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
+        String expectedMessage = String.format(ViewAllCommand.MESSAGE_VIEW_PERSON_DETAILS, list.get(index - 1).getAsTextShowAll());
```
damithc: Line too long?

damithc:
@zzzzwen Yes, you can move common methods to TestUtil class.

test/java/seedu/addressbook/util/TypicalPersons.java
```diff
@@ -0,0 +1,82 @@
...
+            steveSam = new Person(new Name("Steve Sam"),
+                                  new Phone("1231", true),
+                                  new Email("ab@apple.com", true),
+                                  new Address("Silicon Vally", true),
+                                  new UniqueTagList(new Tag("boss")));
+        } catch (IllegalValueException e) {
+            e.printStackTrace();
+            assert false : "not possible";
+        }
+    }
```
damithc: There is another similar class in #120 that might be merged before this. May be you can follow the same? Note how the person names follow a pattern. e.g. First person has initials A B, second one B C, and so on. Even the address follows a pattern and not random. Such patterns help us detect when things go out of place.

zzzzwen: @damithc So I wait for the other pr to merge first then use and edit his `TypicalPersons` for the test?

zzzzwen: Since we both have same class created

damithc: That class is mostly stable now, so you can copy-paste-modify that code into your and continue with your PR. If that PR is merged first, you'll have to update your branch and de-conflict your class a bit at that point.

zzzzwen: ok

zzzzwen:
Updated, I will leave it here first, after merging #120 I will resolve the conflict

zzzzwen:
Updated, ready for review

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,125 @@
...
+    private AddressBook emptyAddressBook;
+    private List<ReadOnlyPerson> emptyDisplayList;
+    private List<ReadOnlyPerson> listWithAll;
+    private List<ReadOnlyPerson> listWithSome;
+
+    @Before
+    public void setUp() throws Exception {
+        TypicalPersons td = new TypicalPersons();
+
+        emptyAddressBook = TestUtil.createAddressBook();
```
damithc: This can be initialized at the same place it's declared. Junit creates a new instance for each test.

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,125 @@
...
+        addressBook = td.getTypicalAddressBook();
```
damithc: Same for this

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+        emptyDisplayList = TestUtil.createList();
+        listWithAll = td.getListWithAllPersons();
+        listWithSome = td.getListWithSomePersons();
```
damithc: On a related note, lines 32, 35, 38, 39, should have been put together as they are connected. i.e. declaring a variable and then using it.

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,125 @@
...
+     */
+    private Command generateViewAllCommand(AddressBook addressBook, List<ReadOnlyPerson> displayList, int index) {
+        Command command = new ViewAllCommand(index);
+        command.setData(addressBook, displayList);
+
+        return command;
+    }
+
+    /**
+     * Asserts that the details person at specific index can be successfully retrieved
```
damithc: `details of the`?

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Asserts that the details of person at specific index cannot be retrieved due to
+     * person not existing in the addressbook.
+     */
+    private void assertViewAllErrorPersonNotInAddressBook(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
+        String expectedMessage = Messages.MESSAGE_PERSON_NOT_IN_ADDRESSBOOK;
+
+        Command command = generateViewAllCommand(addressBook, list, index);
+        assertCommandResult(command, expectedMessage, addressBook, TestUtil.clone(addressBook));
+    }
```
damithc: Can the common part of the above two methods can be extracted as a method?

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,125 @@
...
+     * Asserts that the details of person at specific index cannot be retrieved due to
+     * person not existing in the addressbook.
+     */
+    private void assertViewErrorPersonNotInAddressBook(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
+        String expectedMessage = Messages.MESSAGE_PERSON_NOT_IN_ADDRESSBOOK;
+
+        Command command = generateViewCommand(addressBook, list, index);
+        assertCommandResult(command, expectedMessage, addressBook, TestUtil.clone(addressBook));
+    }
+}
```
damithc: If this class is almost the same as the other test class, we can text both commands in one of the class and make the other class empty with just a comment to explain why it is empty.

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -75,4 +77,18 @@ public static void assertTextFilesEqual(Path path1, Path path2) throws IOExcepti
         List<String> list2 = Files.readAllLines(path2, Charset.defaultCharset());
         assertEquals(String.join("\n", list1), String.join("\n", list2));
     }
+
+    /**
+     * Executes the command, and asserts the result message is as expected.
+     */
+    public static void assertCommandResult(Command command, String expectedMessage, AddressBook addressBook, 
+            AddressBook expectedAddressBook) {
```
damithc: More readable indentation:
```java
public static void assertCommandResult(Command command, String expectedMessage, AddressBook addressBook, 
                                                                         AddressBook expectedAddressBook) {
```

test/java/seedu/addressbook/util/TypicalPersons.java
```diff
@@ -7,47 +10,65 @@
...
+         return list;
+     }
+
+     public List<ReadOnlyPerson> getListWithSomePersons() {
+         List<ReadOnlyPerson> list = getListWithAllPersons();
+         list.remove(0); //remove a person in the list
+         return list;
+     }
+ }
+
```
damithc: Add a line feed at the end of the file

test/java/seedu/addressbook/util/TypicalPersons.java
```diff
@@ -7,47 +10,65 @@
...
-        return new Person[]{amy, bill, candy};
-    }
-
-    public AddressBook getTypicalAddressBook() {
-        AddressBook ab = new AddressBook();
-        loadAddressBookWithSampleData(ab);
-        return ab;
-    }
-
-}
```
damithc: Not sure why this is shown as a diff. Most of it looks identical.

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
zzzzwen: @damithc This `TestUtil.createAddressBook()` is throwing an excpetion in `TestUtil `Class so cannot put outside right?

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
zzzzwen: noted

test/java/seedu/addressbook/util/TypicalPersons.java
zzzzwen: oh, this is because there is an extra whitespace for each line, sorry

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
damithc: In that case can keep.

test/java/seedu/addressbook/commands/ViewAllCommandTest.java
damithc: This comment was intended for the line below. Sorry.

damithc: Also note #134

zzzzwen:
Updated

zzzzwen:
Removed use of TestUtil.createList()

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,152 @@
...
+        assertViewErrorPersonNotInAddressBook(addressBook, listWithAll, listWithAll.size());
+    }
+
+    @Test
+    public void viewCommand_validIndex_returnsPersonDetails() {
+        // person with no private information
+        assertViewSuccess(addressBook, listWithAll, 1);
+
+        // person with some private information
+        assertViewSuccess(addressBook, listWithAll, 2);
```
damithc: You need a person with _all_ info private, to confirm that 'privacy' works for each relevant field.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+        // addressbook has more people than displayList
+        assertViewSuccess(addressBook, listWithSome, 1);
+    }
+
+    /**
+     * Creates a new View command.
```
damithc: `ViewCommand`

damithc: Actually, this comment doesn't add much value. Can be removed.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,152 @@
...
+     * and displayed.
+     */
+    private void assertViewSuccess(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
+        String expectedMessage = String.format(ViewCommand.MESSAGE_VIEW_PERSON_DETAILS,
+                                               list.get(index - 1).getAsTextHidePrivate());
+        assertViewCommandBehaviour(addressBook, list, expectedMessage, index);
+        
+        expectedMessage = String.format(ViewCommand.MESSAGE_VIEW_PERSON_DETAILS,
+                                        list.get(index - 1).getAsTextShowAll());
+        assertViewAllCommandBehaviour(addressBook, list, expectedMessage, index);
```
damithc: `index - 1` appears twice. Assign to local variable?

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Asserts that the details of person at specific index cannot be retrieved due to
+     * invalid index.
+     */
+    private void assertViewErrorInvalidIndex(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
+        String expectedMessage = Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX;
+
+        assertViewCommandBehaviour(addressBook, list, expectedMessage, index);
+        assertViewAllCommandBehaviour(addressBook, list, expectedMessage, index);
```
damithc: Can be simplified as this (after extracting common part of this and next method)?
`assertCommandError(addressBook, list, index, Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX)`

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,152 @@
...
+        String expectedMessage = Messages.MESSAGE_PERSON_NOT_IN_ADDRESSBOOK;
+
+        assertViewCommandBehaviour(addressBook, list, expectedMessage, index);
+        assertViewAllCommandBehaviour(addressBook, list, expectedMessage, index);
+    }
+    
+    /**
+     * Asserts that the View command gives correct feedback information
+     */
+    private void assertViewCommandBehaviour(AddressBook addressBook, List<ReadOnlyPerson> list, String message, int index) {
```
damithc: Reorder parameters so that the three input values come first? 
`message` -> `expectedMessage`

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -75,4 +77,18 @@ public static void assertTextFilesEqual(Path path1, Path path2) throws IOExcepti
         List<String> list2 = Files.readAllLines(path2, Charset.defaultCharset());
         assertEquals(String.join("\n", list1), String.join("\n", list2));
     }
+
+    /**
+     * Executes the command, and asserts the result message is as expected.
+     */
+    public static void assertCommandResult(Command command, String expectedMessage, AddressBook addressBook, 
+                                                                             AddressBook expectedAddressBook) {
```
damithc: Similarly, can reorder parameters so that input comes first and expected values come next.

test/java/seedu/addressbook/util/TypicalPersons.java
```diff
@@ -7,23 +10,27 @@
 import seedu.addressbook.data.person.Name;
 import seedu.addressbook.data.person.Person;
 import seedu.addressbook.data.person.Phone;
+import seedu.addressbook.data.person.ReadOnlyPerson;
+import seedu.addressbook.data.tag.Tag;
 import seedu.addressbook.data.tag.UniqueTagList;
-
+
```
damithc: Remove extra space?

test/java/seedu/addressbook/util/TypicalPersons.java
```diff
@@ -50,4 +57,17 @@ public AddressBook getTypicalAddressBook() {
...
+            list.add(new Person(p));
+        }
+        return list;
+    }
+
+    public List<ReadOnlyPerson> getListWithSomePersons() {
+        List<ReadOnlyPerson> list = getListWithAllPersons();
+        list.remove(0); //remove a person in the list
+        return list;
+    }
```
damithc: This method is very specific to your test class and therefore not a good abstraction. Try something like this instead:
`public List<ReadOnlyPerson> getList(ReadOnlyPerson... personsToInclude)`;

Can be used like this:
`getList(td.amy, td.candy);`

test/java/seedu/addressbook/commands/ViewCommandTest.java
zzzzwen: @damithc In `TypicalPersons` I have a person called `dan` all the information which have the private/public option for him is private.

test/java/seedu/addressbook/commands/ViewCommandTest.java
zzzzwen: I feel that a new local variable, eg, `pos = index - 1` doesn't add much value since `index-1` just refer to the position in the list and won't be changed. Is it necessary to extract to a local variable?

test/java/seedu/addressbook/util/TypicalPersons.java
zzzzwen: noted

zzzzwen:
Updated, but if extract the `assertCommandError` the original will only have 1 line, is this fine

test/java/seedu/addressbook/commands/ViewCommandTest.java
damithc: I meant such a person must be included as one of the test cases in this method. Your comment in line 74 says `some private information` so I guess that's not dan.

test/java/seedu/addressbook/commands/ViewCommandTest.java
damithc: By local, I meant inside the method. Reasons,
1. It's not good when a calculation is duplicated. 
2. its meaning is not immediately clear. 

Something like this will help clear both problems.

`int internalIndex = index - 1;  // because index is one-indexed`

damithc:
@zzzzwen 
> if extract the assertCommandError the original will only have 1 line, is this fine?

Yes, it is fine to have one line methods. The purpose of a method can be simply to call another method with
some specific parameters.

test/java/seedu/addressbook/commands/ViewCommandTest.java
zzzzwen: oh, I get it

zzzzwen:
Updated

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,153 @@
...
+     * Asserts that the details of the person at specific index can be successfully retrieved
+     * and displayed.
+     */
+    private void assertViewSuccess(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
+        String expectedMessage = String.format(ViewCommand.MESSAGE_VIEW_PERSON_DETAILS,
+                                               list.get(index - 1).getAsTextHidePrivate());
+        assertViewCommandBehaviour(addressBook, list, index, expectedMessage);
+
+        expectedMessage = String.format(ViewCommand.MESSAGE_VIEW_PERSON_DETAILS,
+                                        list.get(index - 1).getAsTextShowAll());
```
damithc: Further to my previous comment about `index-1`, in fact `list.get(index - 1)` can be assigned to a local variable.
`ReadOnlyPerson personBeingViewed = list.get(index-1); // -1 because index is one-indexed`

zzzzwen:
Updated. Noted the use of local variable and constant from this and previous commits. Thanks for detailed explanation

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -39,7 +41,7 @@ public static AddressBook createAddressBook(Person... persons) throws DuplicateP
     /**
      * Creates a list of persons.
      */
-    public static List<ReadOnlyPerson> createList(Person...persons) {
+    public static List<ReadOnlyPerson> createList(Person... persons) {
```
damithc: Not related to this PR? In that case, exclude it.

test/java/seedu/addressbook/util/TypicalPersons.java
```diff
@@ -50,4 +57,19 @@ public AddressBook getTypicalAddressBook() {
         return ab;
     }
 
+    public List<ReadOnlyPerson> getListWithAllPersons() {
```
damithc: Also, not sure if you need this method. Can you use `getTypicalPersons()` instead, and convert it to a list using Arrays.asList()?

zzzzwen:
Updated

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,157 @@
...
+    private Command generateViewAllCommand(AddressBook addressBook, List<ReadOnlyPerson> displayList, int index) {
+        Command command = new ViewAllCommand(index);
+        command.setData(addressBook, displayList);
+
+        return command;
+    }
+
+    /**
+     * Asserts that the details of the person at specific index can be successfully retrieved
+     * and displayed.
```
damithc: `Asserts that both a ViewCommand and a ViewAllCommand can retrieve from the {@code addressBook} details of the person at the given {@code index} in the given {@code list}.`

damithc: Also add
`@param index one-indexed position of the target person in the list`

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,157 @@
...
+    /**
+     * Asserts that the details of person at specific index cannot be retrieved due to
+     * person not existing in the addressbook.
+     */
+    private void assertViewErrorPersonNotInAddressBook(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
+        assertCommandError(addressBook, list, index, Messages.MESSAGE_PERSON_NOT_IN_ADDRESSBOOK);
+    }
+
+    /**
+     * Asserts that the Viewcommand and ViewAllcommand reports error for given input
```
damithc: `... reports the given error for the given ...`

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,157 @@
...
+     */
+    private void assertCommandError(AddressBook addressBook, List<ReadOnlyPerson> list, int index,
+                                                                            String expectedMessage) {
+        assertViewCommandBehaviour(addressBook, list, index, expectedMessage);
+        assertViewAllCommandBehaviour(addressBook, list, index, expectedMessage);
+    }
+
+    /**
+     * Asserts that the View command gives correct feedback information
```
damithc: `correct` --> `given`

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+import seedu.addressbook.data.person.Phone;
+import seedu.addressbook.data.person.ReadOnlyPerson;
+import seedu.addressbook.data.tag.UniqueTagList;
+import seedu.addressbook.util.TestUtil;
+import seedu.addressbook.util.TypicalPersons;
+
+import static seedu.addressbook.util.TestUtil.assertCommandResult;
+
+public class ViewCommandTest {
+    private AddressBook addressBook;
```
pyokagan: Hmph, we can't use `TestUtil.createAddressBook()` here because it throws a checked exception. However, if we step back a little, we see that `TestUtil.createAddressBook()` should *not* be passing on the `DuplicatePersonException` because it's *only used for test data*. (i.e. the same reason why `TypicalPersons` just `asserts false` if it encounters an `IllegalValueException`.)

So, I think you should just add this in to your PR:
```diff
diff --git a/test/java/seedu/addressbook/util/TestUtil.java b/test/java/seedu/addressbook/util/TestUtil.java
index 8f4eaf8..a1f0836 100644
--- a/test/java/seedu/addressbook/util/TestUtil.java
+++ b/test/java/seedu/addressbook/util/TestUtil.java
@@ -26,13 +26,16 @@ import seedu.addressbook.data.tag.UniqueTagList;
 public class TestUtil {
     /**
      * Creates an address book containing the given persons.
-     * @throws DuplicatePersonException if two or more given persons are the same
      */
-    public static AddressBook createAddressBook(Person... persons) throws DuplicatePersonException {
+    public static AddressBook createAddressBook(Person... persons) {
         AddressBook addressBook = new AddressBook();
 
         for (Person person : persons) {
-            addressBook.addPerson(person);
+            try {
+                addressBook.addPerson(person);
+            } catch (DuplicatePersonException e) {
+                throw new AssertionError(e);
+            }
         }
 
         return addressBook;
```

No point letting the quality of your PR suffer just because the code base just happens to be like this already.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+    private AddressBook emptyAddressBook;
+    private List<ReadOnlyPerson> emptyDisplayList = new ArrayList<ReadOnlyPerson>();
+    private List<ReadOnlyPerson> listWithAll;
```
pyokagan: May be useful to name this lists so that first-time readers will know what they will be used for.

e.g. `displayListWithAllTypicalPersons`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+    private List<ReadOnlyPerson> listWithSome;
+
+    @Before
+    public void setUp() throws Exception {
+        emptyAddressBook = TestUtil.createAddressBook();
+
+        TypicalPersons td = new TypicalPersons();
+        addressBook = td.getTypicalAddressBook();
+        listWithAll = Arrays.asList(td.getTypicalPersons());
+        listWithSome = Arrays.asList(td.amy, td.candy, td.dan);
```
pyokagan: All of these *should* be moved out of the `setUp()` method.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+        // person with all private information
+        assertViewSuccess(addressBook, listWithAll, 4);
+
+        // addressbook has more people than displayList
+        // This can happen when view from filtered list caused by some commands(eg. FindCommand)
+        assertViewSuccess(addressBook, listWithSome, 1);
+    }
+
+    private Command generateViewCommand(AddressBook addressBook, List<ReadOnlyPerson> displayList, int index) {
```
pyokagan: Why `displayList`? Isn't this naming inconsistent with what `Command` already uses? i.e. `relevantPersons`?

If you prefer the name `displayList`, then you should change all instances of `relevantPersons` to `displayList` as well.

pyokagan: Also, `index` is also inconsistent with the naming which `ViewCommand` or `Command` uses. i.e. `targetIndex`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
pyokagan: Also, maybe name this `typicalAddressBook` to signal that it is filled with all the `TypicalPersons`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+        emptyAddressBook = TestUtil.createAddressBook();
+
+        TypicalPersons td = new TypicalPersons();
+        addressBook = td.getTypicalAddressBook();
+        listWithAll = Arrays.asList(td.getTypicalPersons());
+        listWithSome = Arrays.asList(td.amy, td.candy, td.dan);
+    }
+
+    @Test
+    public void viewCommand_invalidIndex_returnsInvalidIndexMessage() {
```
pyokagan: @damithc Do you prefer the naming to be `execute_invalidIndex...` or `viewCommand_invalidIndex...`?

I see the `viewCommand_...` naming was introduced by #118 (my fault, I should have noticed it), which is probably at odds with the project's naming convention.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+    @Test
+    public void viewCommand_personNotInAddressBook_returnsPersonNotInAddressBookMessage() throws Exception {
+        // generate person not in addressbook, add to displayList
+        ReadOnlyPerson someone = new Person(new Name("me"),
+                                            new Phone("123", true),
+                                            new Email("some@hey.go", true),
+                                            new Address("nus", false),
+                                            new UniqueTagList(Collections.emptySet()));
+        listWithAll = new ArrayList<ReadOnlyPerson>(listWithAll);
+        listWithAll.add(someone);
```
pyokagan: This goes against how `listWithAll` is used in other tests (where `listWithAll` is filled with all the typical persons).

I think you should just define a local variable for this.

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -75,4 +77,18 @@ public static void assertTextFilesEqual(Path path1, Path path2) throws IOExcepti
         List<String> list2 = Files.readAllLines(path2, Charset.defaultCharset());
         assertEquals(String.join("\n", list1), String.join("\n", list2));
     }
+
+    /**
+     * Executes the command, and asserts the result message is as expected.
+     */
+    public static void assertCommandResult(Command command, AddressBook addressBook, AddressBook expectedAddressBook,
+                                                                                               String expectedMessage) {
```
pyokagan: Besides the line wrapping looking funny, this actually still exceeds 110 characters.

pyokagan: @damithc So, I do wonder whether this method is doing too much to be considered a "utility" method. Of the four arguments, the 1st is used in the 1st statement, the 4th is used in the 2nd statement, and the 2nd and 3rd are used in the 3rd statement. Note how the use of these arguments do not overlap at all.

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -75,4 +77,18 @@ public static void assertTextFilesEqual(Path path1, Path path2) throws IOExcepti
...
+        CommandResult result = command.execute();
+
+        // asserts the result message is correct as expected
+        assertEquals(expectedMessage, result.feedbackToUser);
```
pyokagan: Should we be comparing the `CommandResult` instead, rather than just one field of it?

For example, we may wish to check that `CommandResult.getRelevantPersons()` returns `Optional.empty()` or a list that we expect.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+        expectedMessage = String.format(ViewCommand.MESSAGE_VIEW_PERSON_DETAILS,
+                                        personToBeViewed.getAsTextShowAll());
+        assertViewAllCommandBehaviour(addressBook, list, index, expectedMessage);
+    }
+
+    /**
+     * Asserts that the details of person at specific index cannot be retrieved due to
+     * invalid index.
+     */
+    private void assertViewErrorInvalidIndex(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
```
pyokagan: Does this really need its own method?

Yes, I know the alternative is repeating `Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX` in `viewCommand_invalidIndex_returnsInvalidIndexMessage()`, but then again with this method you still need to repeat `assertViewErrorInvalidIndex` anyway, which probably means this method doesn't add much value.

If the reason is because `Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX` is too long, then you can assign in to a variable with a shorter name.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+        assertCommandError(addressBook, list, index, Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
+    }
+
+    /**
+     * Asserts that the details of person at specific index cannot be retrieved due to
+     * person not existing in the addressbook.
+     */
+    private void assertViewErrorPersonNotInAddressBook(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
```
pyokagan: Does this really need its own method? Same reasoning as `assertViewErrorInvalidIndex`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,159 @@
...
+    private void assertCommandError(AddressBook addressBook, List<ReadOnlyPerson> list, int index,
+                                                                            String expectedMessage) {
+        assertViewCommandBehaviour(addressBook, list, index, expectedMessage);
+        assertViewAllCommandBehaviour(addressBook, list, index, expectedMessage);
+    }
+
+    /**
+     * Asserts that the View command gives given feedback information
+     */
+    private void assertViewCommandBehaviour(AddressBook addressBook, List<ReadOnlyPerson> list, int index, String expectedMessage) {
```
pyokagan: @damithc Do you prefer American or British spelling? :-P

damithc: > Do you prefer American or British spelling? :-P

Um, never thought about it before. Can we allow either one?

test/java/seedu/addressbook/util/TestUtil.java
damithc: > Should we be comparing the CommandResult instead, rather than just one field of it?

Yes, that's actually better. I don't mind leaving it as a TODO though.

test/java/seedu/addressbook/commands/ViewCommandTest.java
zzzzwen: If inline this method and put Message into the test case, does it really make sense or help improve readability?

eg. `assertViewErrorPersonNotInAddressBook(emptyAddressBook, listWithAll, 1)` -> `assertCommandError(emptyAddressBook, listWithAll, 1, Message.PERSON_NOT_IN_ADDRESSBOOK)`

zzzzwen: `assertCommandError` does not imply what expected error is and can we assume that reader can infer from the parameter about the expected error? In this case, `assertViewErrorInvalidIndex` is clearer for me

test/java/seedu/addressbook/util/TestUtil.java
zzzzwen: @pyokagan in this case, how to indent for the parameters

zzzzwen: can you show me some ideas?

test/java/seedu/addressbook/commands/ViewCommandTest.java
zzzzwen: those need `TypicalPersons td = new TypicalPersons();` can be moved outside?

pyokagan: @zzzzwen 
> those need TypicalPersons td = new TypicalPersons(); can be moved outside?

Yes.

test/java/seedu/addressbook/commands/ViewCommandTest.java
pyokagan: @zzzzwen 
> assertCommandError does not imply what expected error is and can we assume that reader can infer from the parameter about the expected error

Doesn't your test name already say `_returnsInvalidIndexMessage`?

> In this case, assertViewErrorInvalidIndex is clearer for me

OK then.

test/java/seedu/addressbook/commands/ViewCommandTest.java
pyokagan: @damithc 
> Can we allow either one?

I'd recommend not, otherwise we will get inconsistent naming, which lowers discoverability.

I'd would also recommend going with American spelling, as I think that's what the Java standard library uses, and probably what our Java dependencies will also use as well.

test/java/seedu/addressbook/util/TestUtil.java
pyokagan: @zzzzwen 
> in this case, how to indent for the parameters

One method you can try is to split the argument list at a different location instead, e.g.
```java
    public static void assertCommandResult(Command command, AddressBook addressBook,
                                           AddressBook expectedAddressBook, String expectedMessage) {
```

zzzzwen:
Updated

test/java/seedu/addressbook/commands/ViewCommandTest.java
damithc: >I'd would also recommend going with American spelling

@pyokagan  Ok, will add to oss-generic/process
@zzzzwen change to `Behavior`

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -26,13 +26,17 @@
...
-    public static AddressBook createAddressBook(Person... persons) throws DuplicatePersonException {
+    public static AddressBook createAddressBook(Person... persons) {
         AddressBook addressBook = new AddressBook();
 
         for (Person person : persons) {
-            addressBook.addPerson(person);
+            try {
+                addressBook.addPerson(person);
+            } catch (DuplicatePersonException e) {
+                // TODO Auto-generated catch block
```
damithc: ?

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -75,4 +80,19 @@ public static void assertTextFilesEqual(Path path1, Path path2) throws IOExcepti
         List<String> list2 = Files.readAllLines(path2, Charset.defaultCharset());
         assertEquals(String.join("\n", list1), String.join("\n", list2));
     }
+
+    /**
+     * Executes the command, and asserts the result message is as expected.
+     */
+    public static void assertCommandResult(Command command, AddressBook addressBook,
+                                           AddressBook expectedAddressBook, String expectedMessage) {
```
pyokagan: Problems with this method:

1. Naming: `CommandResult` is a class, but `assertCommandResult` does more than check for `CommandResult`. Maybe this should be called `assertCommandBehavior` instead?

2. Not watertight: it assumes that the passed-in `addressBook` is the address book that has been registered with the command. Which may not be true, and opens the door to false negatives.

3. It does not check for the `getRelevantPersons()` of the returned `CommandResult`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private Command generateViewCommand(AddressBook addressBook, List<ReadOnlyPerson> relevantPersons, int targetIndex) {
+        Command command = new ViewCommand(targetIndex);
+        command.setData(addressBook, relevantPersons);
+
+        return command;
+    }
+
+    private Command generateViewAllCommand(AddressBook addressBook, List<ReadOnlyPerson> relevantPersons, int targetIndex) {
```
pyokagan: The code of `generateViewCommand()` and `generateViewAllCommand()` look very similar, which hints that they can be combined into a more generic method.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,161 @@
...
+    private void assertViewCommandBehaviour(AddressBook addressBook, List<ReadOnlyPerson> list, int index, 
+                                            String expectedMessage) {
+        Command command = generateViewCommand(addressBook, list, index);
+        assertCommandResult(command, addressBook, TestUtil.clone(addressBook), expectedMessage);
+    }
+
+    /**
+     * Asserts that the ViewAll command gives given feedback information
+     */
+    private void assertViewAllCommandBehaviour(AddressBook addressBook, List<ReadOnlyPerson> list, int index,
```
pyokagan: `assertViewCommandBehaviour()` and `assertViewCommandBehaviour()` look very similar, which hints they can be combined into a more generic method.

By the way, these methods are still using British spelling.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,161 @@
...
+    private List<ReadOnlyPerson> listWithSome;
+
+    @Before
+    public void setUp() {
+        emptyAddressBook = TestUtil.createAddressBook();
+
+        TypicalPersons td = new TypicalPersons();
+        typicalAddressBook = td.getTypicalAddressBook();
+        displayListWithAllTypicalPersons = Arrays.asList(td.getTypicalPersons());
+        listWithSome = Arrays.asList(td.amy, td.candy, td.dan);
```
pyokagan: These can be moved out of `setUp()`.

pyokagan:
@zzzzwen 

There are quite a lot of repetitive utility methods in `ViewCommandTest`. Having too many methods which call each other in a class makes it hard to read because readers have to jump all over the code in order to understand what a test case does. For a more exaggerated version of the problem, see the code of AddressBook-Level1.

One way of getting rid of the repetitive `generateViewCommand()`, `generateViewAllCommand()`, `assertViewCommandBehavior()`, `assertViewAllCommandBehavior()` is to use a utility method that accepts a `Command` directly instead:

```java
    private static void assertViewBehavior(Command viewCommand, AddressBook addressBook,
                                           List<ReadOnlyPerson> list, String expectedMessage) {
        AddressBook expectedAddressBook = TestUtil.clone(addressBook);

        viewCommand.setData(addressBook, list);
        CommandResult result = viewCommand.execute();

        // Feedback message is as expected and there are no relevant persons returned
        assertEquals(expectedMessage, result.feedbackToUser);
        assertEquals(Optional.empty(), result.getRelevantPersons());

        // Address book was not modified
        assertEquals(expectedAddressBook.getAllPersons(), addressBook.getAllPersons());
    }
```

`assertViewSuccess()` and `assertViewError()` can then be written as follows:
```java
    private void assertViewSuccess(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
        ReadOnlyPerson personToBeViewed = list.get(index - 1); // -1 is because index is one-indexed

        String expectedMessage = String.format(ViewCommand.MESSAGE_VIEW_PERSON_DETAILS,
                                               personToBeViewed.getAsTextHidePrivate());
        assertViewBehavior(new ViewCommand(index), addressBook, list, expectedMessage);

        expectedMessage = String.format(ViewCommand.MESSAGE_VIEW_PERSON_DETAILS,
                                        personToBeViewed.getAsTextShowAll());
        assertViewBehavior(new ViewAllCommand(index), addressBook, list, expectedMessage);
    }

    private static void assertViewError(AddressBook addressBook, List<ReadOnlyPerson> displayList,
                                                                int index, String expectedMessage) {
        assertViewBehavior(new ViewCommand(index), addressBook, displayList, expectedMessage);
        assertViewBehavior(new ViewAllCommand(index), addressBook, displayList, expectedMessage);
    }
```

This reduces your *eight* utility methods down to only *three*.

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -24,13 +26,16 @@
...
-    public static AddressBook createAddressBook(Person... persons) throws DuplicatePersonException {
+    public static AddressBook createAddressBook(Person... persons) {
         AddressBook addressBook = new AddressBook();
 
         for (Person person : persons) {
-            addressBook.addPerson(person);
+            try {
+                addressBook.addPerson(person);
+            } catch (DuplicatePersonException e) {
+                e.printStackTrace();
```
pyokagan: Shouldn't you throw an error rather than just swallowing it? Having duplicate persons in test data is definitely a programmer error.

test/java/seedu/addressbook/commands/ViewCommandTest.java
zzzzwen: @pyokagan but it needs `td` the `TypicalPersons` instance to get the typical persons, if moved outside, needs to change all the relevant persons and methods to static?

zzzzwen: ok

test/java/seedu/addressbook/util/TestUtil.java
zzzzwen: @pyokagan there is a `TODO` below which asks to implement equals method in `CommandResult`, then compare the expected `CommandResult `and actual

zzzzwen: but in this case why there is false negative? previously I passed in the `addressbook` after executing the command. Then how to check the false negatives

zzzzwen:
Updated, but ViewCommand and ViewAllCommand are two separate commands and my original idea is to split them into two different blocks of methods to show they are different.

test/java/seedu/addressbook/util/TestUtil.java
pyokagan: @zzzzwen 
> there is a TODO below which asks to implement equals method in CommandResult, then compare the expected CommandResult and actual

The point is that the code is not checking that the `CommandResult` returned by `ViewCommand` and `ViewAllCommand` has `getRelevantPersons() == Optional.empty()`.

> but in this case why there is false negative?

Let's say that we have `AddressBook` `a` and `b`, and a `ViewCommand` called `command`. We do `command.setData(a, ...)` but then pass `b` into `assertCommandResult()`, e.g. `assertCommandResult(command, b, TestUtil.clone(b), ...)`.

Now, let's say that `command` has a bug that ends up modifying the address book `a`. However, with the current implementation of `assertCommandResult()`, the test will still pass since `b` was not modified, even though `command` has a bug.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,139 @@
...
+                                                                int targetVisibleIndex, String expectedMessage) {
+        assertViewBehavior(new ViewCommand(targetVisibleIndex), addressBook, displayList, expectedMessage);
+        assertViewBehavior(new ViewAllCommand(targetVisibleIndex), addressBook, displayList, expectedMessage);
+    }
+
+    /**
+     * Asserts that the ViewCommand and ViewAllCommand show given feedback information
+     */
+    private static void assertViewBehavior(Command viewCommand, AddressBook addressBook,
+        List<ReadOnlyPerson> list, String expectedMessage) {
```
pyokagan: Indent this further to the right.

test/java/seedu/addressbook/util/TestUtil.java
```diff
@@ -75,4 +78,5 @@ public static void assertTextFilesEqual(Path path1, Path path2) throws IOExcepti
         List<String> list2 = Files.readAllLines(path2, Charset.defaultCharset());
         assertEquals(String.join("\n", list1), String.join("\n", list2));
     }
+
```
pyokagan: Unnecessary change.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,139 @@
...
+import seedu.addressbook.data.person.Person;
+import seedu.addressbook.data.person.Phone;
+import seedu.addressbook.data.person.ReadOnlyPerson;
+import seedu.addressbook.data.tag.UniqueTagList;
+import seedu.addressbook.util.TestUtil;
+import seedu.addressbook.util.TypicalPersons;
+
+public class ViewCommandTest {
+    private TypicalPersons td = new TypicalPersons();
+
```
pyokagan: Trailing whitespace

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,139 @@
...
+    @Test
+    public void execute_invalidIndex_returnsInvalidIndexMessage() {
+        // empty addressbook
+        assertViewErrorInvalidIndex(emptyAddressBook, emptyDisplayList, 1);
+
+        // non-empty addressbook
+        assertViewErrorInvalidIndex(typicalAddressBook, displayListWithAllTypicalPersons, -1);
+        assertViewErrorInvalidIndex(typicalAddressBook, displayListWithAllTypicalPersons, 0);
+        assertViewErrorInvalidIndex(typicalAddressBook, displayListWithAllTypicalPersons, displayListWithAllTypicalPersons.size() + 1);
```
pyokagan: Overlong line way above 110 chars.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Test
+    public void execute_personNotInAddressBook_returnsPersonNotInAddressBookMessage() throws Exception {
+        // generate person not in addressbook, add to displayList
+        ReadOnlyPerson someone = new Person(new Name("me"),
+                                            new Phone("123", true),
+                                            new Email("some@hey.go", true),
+                                            new Address("nus", false),
+                                            new UniqueTagList(Collections.emptySet()));
+        List <ReadOnlyPerson> displayListWithExtraPerson = new ArrayList<ReadOnlyPerson>(displayListWithAllTypicalPersons);
```
pyokagan: Overlong line way above 110 chars.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,139 @@
...
+        displayListWithExtraPerson.add(someone);
+
+        // empty addressbook
+        assertViewErrorPersonNotInAddressBook(emptyAddressBook, displayListWithExtraPerson, 1);
+
+        // non-empty addressbook
+        assertViewErrorPersonNotInAddressBook(typicalAddressBook, displayListWithExtraPerson, displayListWithExtraPerson.size());
```
pyokagan: Overlong line way above 110 chars.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,139 @@
...
+     * person not existing in the addressbook.
+     */
+    private void assertViewErrorPersonNotInAddressBook(AddressBook addressBook, List<ReadOnlyPerson> list, int index) {
+        assertViewError(addressBook, list, index, Messages.MESSAGE_PERSON_NOT_IN_ADDRESSBOOK);
+    }
+
+    /**
+     * Asserts that both a ViewCommand and a ViewAllCommand can retrieve from
+     * the {@code addressBook} details of the person at the given {@code index} in the given {@code list}.
+     *
```
pyokagan: Trailing whitespace.

pyokagan:
@zzzzwen I notice you are still using the `displayList` terminology in your code. As I [noted previously](https://github.com/se-edu/addressbook-level2/pull/131#discussion_r94742718), this is inconsistent with the terminology already in use (`relevantPersons`).

Maybe `displayList` is a better name ( I personally think that `relevantPersons` is not such a great term, but I'm not so sure that `displayList` is a good replacement either. @damithc What do you think?  ) But if so, you need to commit to changing all instances of `relevantPersons` in the code base to use your `displayList` terminology. Are you up to the task? If so, make sure you open up an issue for that.

zzzzwen:
Configured and updated. Replaced all with parameters in `Command`

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,146 @@
...
+import seedu.addressbook.data.tag.UniqueTagList;
+import seedu.addressbook.util.TestUtil;
+import seedu.addressbook.util.TypicalPersons;
+
+public class ViewCommandTest {
+    private TypicalPersons td = new TypicalPersons();
+
+    private AddressBook typicalAddressBook = td.getTypicalAddressBook();
+    private AddressBook emptyAddressBook = TestUtil.createAddressBook();
+    private List<ReadOnlyPerson> emptyDisplayList = Collections.emptyList();
```
pyokagan: Maybe rename `emptyDisplayList` to `emptyPersonList`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,146 @@
...
+    private List<ReadOnlyPerson> displayListWithAllTypicalPersons = Arrays.asList(td.getTypicalPersons());
```
pyokagan: Maybe rename `displayListWithAllTypicalPersons` to `allTypicalPersonsList` or `listWithAllTypicalPersons`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,146 @@
...
+    private List<ReadOnlyPerson> listWithSome = Arrays.asList(td.amy, td.candy, td.dan);
```
pyokagan: Maybe rename `listWithSome` to `listWithSomeTypicalPersons`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Test
+    public void execute_personNotInAddressBook_returnsPersonNotInAddressBookMessage() throws Exception {
+        // generate person not in addressbook, add to displayList
+        ReadOnlyPerson someone = new Person(new Name("me"),
+                                            new Phone("123", true),
+                                            new Email("some@hey.go", true),
+                                            new Address("nus", false),
+                                            new UniqueTagList(Collections.emptySet()));
+        List <ReadOnlyPerson> displayListWithExtraPerson
```
pyokagan: Can just be `listWithExtraPerson`. Also, remove the space in `List <ReadOnlyPerson>`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,146 @@
...
+        // non-empty addressbook
+        assertViewErrorInvalidIndex(typicalAddressBook, displayListWithAllTypicalPersons, -1);
+        assertViewErrorInvalidIndex(typicalAddressBook, displayListWithAllTypicalPersons, 0);
+        assertViewErrorInvalidIndex(typicalAddressBook, displayListWithAllTypicalPersons,
+                displayListWithAllTypicalPersons.size() + 1);
+    }
+
+    @Test
+    public void execute_personNotInAddressBook_returnsPersonNotInAddressBookMessage() throws Exception {
+        // generate person not in addressbook, add to displayList
```
pyokagan: More descriptive: `generate list with person not in addressbook`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,146 @@
...
+        // addressbook has more people than displayList
+        // This can happen when view from filtered list caused by some commands(eg. FindCommand)
+        assertViewSuccess(typicalAddressBook, listWithSome, 1);
+    }
+
+    /**
+     * Asserts that the details of person at specific index cannot be retrieved due to
+     * invalid index.
+     */
+    private void assertViewErrorInvalidIndex(AddressBook addressBook, List<ReadOnlyPerson> list,
```
pyokagan: Rename `list` to `relevantPersons`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,146 @@
...
+                                             int targetVisibleIndex) {
+        assertViewError(addressBook, list, targetVisibleIndex, Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
+    }
+
+    /**
+     * Asserts that the details of person at specific index cannot be retrieved due to
+     * person not existing in the addressbook.
+     */
+    private void assertViewErrorPersonNotInAddressBook(AddressBook addressBook, List<ReadOnlyPerson> list,
```
pyokagan: Rename `list` to `relevantPersons`.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,146 @@
...
+     * Asserts that the Viewcommand and ViewAllcommand reports the given error for the given input
+     */
+    private static void assertViewError(AddressBook addressBook, List<ReadOnlyPerson> relevantPersons,
+                                        int targetVisibleIndex, String expectedMessage) {
+        assertViewBehavior(new ViewCommand(targetVisibleIndex), addressBook, relevantPersons, expectedMessage);
+        assertViewBehavior(new ViewAllCommand(targetVisibleIndex), addressBook, relevantPersons, expectedMessage);
+    }
+
+    /**
+     * Asserts that the ViewCommand and ViewAllCommand show given feedback information
```
pyokagan: It does not just check for feedback information. You may wish to enumerate the list of things it checks for, e.g.
```
Checks that viewCommand exhibits the correct view command behavior, namely:

1. The feedback message of the CommandResult it returns matches expectedMessage
2. The CommandResult it returns has no relevant persons.
3. ...
```

zzzzwen:
Updated, the `displayList` in the `assertViewErrorInvalidIndex` and `assertViewErrorPersonNotInAddressBook` are not changed in the previous commits because the terminology `relevantPersons` is for constructing the `Command` and those two in the methods still don't touch the `Command` yet, so I didn't change them in the two methods.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,150 @@
...
+     * person not existing in the addressbook.
+     */
+    private void assertViewErrorPersonNotInAddressBook(AddressBook addressBook, List<ReadOnlyPerson> relevantPersons,
+                                                       int targetVisibleIndex) {
+        assertViewError(addressBook, relevantPersons, targetVisibleIndex, Messages.MESSAGE_PERSON_NOT_IN_ADDRESSBOOK);
+    }
+
+    /**
+     * Asserts that both a ViewCommand and a ViewAllCommand can retrieve from
+     * the {@code addressBook} details of the person at the given {@code index} in the given {@code list}.
```
pyokagan: Change `{@code index}` to `{@code targetVisibleIndex}`?

Change `{@code list}` to `{@code relevantPersons} list`?

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,150 @@
...
+        assertViewSuccess(typicalAddressBook, listWithSomeTypicalPersons, 1);
+    }
+
+    /**
+     * Asserts that the details of person at specific index cannot be retrieved due to
+     * invalid index.
+     */
+    private void assertViewErrorInvalidIndex(AddressBook addressBook, List<ReadOnlyPerson> relevantPersons,
+                                             int targetVisibleIndex) {
+        assertViewError(addressBook, relevantPersons, targetVisibleIndex, Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
```
pyokagan: This line is much longer than 110 chars.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Asserts that both a ViewCommand and a ViewAllCommand can retrieve from
+     * the {@code addressBook} details of the person at the given {@code index} in the given {@code list}.
+     *
+     * @param targetVisibleIndex one-indexed position of the target person in the list
+     */
+    private void assertViewSuccess(AddressBook addressBook, List<ReadOnlyPerson> relevantPersons,
+                                   int targetVisibleIndex) {
+        ReadOnlyPerson personToBeViewed = relevantPersons.get(targetVisibleIndex - 1); // -1 is because index is one-indexed
```
pyokagan: This line is much longer than 110 chars.

pyokagan: We could put it like that:
```java
        // -1 is because index is one-indexed
        ReadOnlyPerson personToBeViewed = relevantPersons.get(targetVisibleIndex - 1);
```
but imo that looks weird because usually a comment at the code block describes what the code below overall does, but in this case all we are doing is just commenting on the use of `-1`.

So maybe we need to add a description on what the code below does, even though it may be obvious. So something like:
```java
        // Get person to be viewed (targetVisibleIndex - 1 because index is one-indexed)
        ReadOnlyPerson personToBeViewed = relevantPersons.get(targetVisibleIndex - 1);
```

zzzzwen:
Updated

damithc:
@zzzzwen the `Fixes #...` info is missing from the PR description.

zzzzwen:
oh, added

pyokagan:
@damithc Can reconfirm your approval? PR has changed since your last review.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,154 @@
...
+    @Test
+    public void execute_invalidIndex_returnsInvalidIndexMessage() {
+        // empty addressbook
+        assertViewErrorInvalidIndex(emptyAddressBook, emptyPersonList, 1);
+
+        // non-empty addressbook
+        assertViewErrorInvalidIndex(typicalAddressBook, listWithAllTypicalPersons, -1);
+        assertViewErrorInvalidIndex(typicalAddressBook, listWithAllTypicalPersons, 0);
+        assertViewErrorInvalidIndex(typicalAddressBook, listWithAllTypicalPersons,
+                listWithAllTypicalPersons.size() + 1);
```
damithc: Minor nit: indent for better readability. This way, `(` and `)` that demarcate the parameter list appear on far left and far right of the parameters, respectively.

```java
 assertViewErrorInvalidIndex(typicalAddressBook, listWithAllTypicalPersons,
                                       listWithAllTypicalPersons.size() + 1);
```
This applies to a few other places as well.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Test
+    public void execute_personNotInAddressBook_returnsPersonNotInAddressBookMessage() throws Exception {
+        // generate list with person not in addressbook, add to list
+        ReadOnlyPerson someone = new Person(new Name("me"),
```
damithc: `someone` --> `stranger`?

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,154 @@
...
+        assertViewSuccess(typicalAddressBook, listWithAllTypicalPersons, 1);
+
+        // person with some private information
+        assertViewSuccess(typicalAddressBook, listWithAllTypicalPersons, 2);
+
+        // person with all private information
+        assertViewSuccess(typicalAddressBook, listWithAllTypicalPersons, 4);
+
+        // addressbook has more people than the list
+        // This can happen when view from filtered list caused by some commands(eg. FindCommand)
```
damithc: This sentence is not very clear.
`This can happen when a command causes the list to show only a sub-set of persons (e.g. FindCommand).`

Also, full sentences should end with full stop.

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Asserts that both a ViewCommand and a ViewAllCommand can retrieve from
+     * the {@code addressBook} details of the person at the given {@code targetVisibleIndex}
+     * in the given {@code relevantPersons} list.
+     *
+     * @param targetVisibleIndex one-indexed position of the target person in the list
+     */
+    private void assertViewSuccess(AddressBook addressBook, List<ReadOnlyPerson> relevantPersons,
+                                   int targetVisibleIndex) {
+        // Get person to be viewed (targetVisibleIndex - 1 because index is one-indexed)
```
damithc: `get` (to be consistent with other comments in this class)?

zzzzwen:
Updated

test/java/seedu/addressbook/commands/ViewCommandTest.java
```diff
@@ -74,8 +74,8 @@ public void execute_validIndex_returnsPersonDetails() {
         // person with all private information
         assertViewSuccess(typicalAddressBook, listWithAllTypicalPersons, 4);
 
-        // addressbook has more people than the list
-        // This can happen when view from filtered list caused by some commands(eg. FindCommand)
+        // addressbook has more people than the list.
+        // this can happen when a command causes the list to show only a sub-set of persons(e.g. FindCommand).
```
damithc: Full sentences should start with a upper case letter too. :-) So it should either a full sentence starting with a upper case letter and ending with a full stop, or just a short phrase starting with a lower case letter and no full stop at the end.

zzzzwen:
ok

pyokagan:
@damithc Any more comments for this PR?

Issue #130, opened 2016-12-26
Change file extension requirement from .txt to .xml


damithc:
.txt is a legacy requirement from level 1.
This address book saves data in xml format, so the file should have the extension .xml

luyangkenneth:
Hello! I believe this issue can be closed :)

Issue #117, opened 2016-12-07
Names with unicode characters are treated as invalid


Neurrone:
Noticed this happening in the last 2 test cases in input.txt. Think it might be caused by a letter being [represented with 2 code points](http://www.regular-expressions.info/unicode.html), in which case \p{L} wouldn't match.

pyokagan:
I can reproduce on Windows only. `runtests.bat` fails for me.

```
Comparing files actual.txt and EXPECTED.TXT
***** actual.txt
|| Enter command: || [Command entered:  add Björn Borg p/98765432 e/borg@gmail.com a/311, Clementi Ave 2, #02-25]
|| Person names should be spaces or alphabetic characters
|| ===================================================
***** EXPECTED.TXT
|| Enter command: || [Command entered:  add Björn Borg p/98765432 e/borg@gmail.com a/311, Clementi Ave 2, #02-25]
|| New person added: Björn Borg Phone: 98765432 Email: borg@gmail.com Address: 311, Clementi Ave 2, #02-25 Tags: 
|| ===================================================
*****

***** actual.txt
||      3. John-Doe jr. Phone: 98765432 Email: johnd@gmail.com Address: 311, Clementi Ave 2, #02-25 Tags: 
|| 
|| 3 persons listed!
|| ===================================================
***** EXPECTED.TXT
||      3. John-Doe jr. Phone: 98765432 Email: johnd@gmail.com Address: 311, Clementi Ave 2, #02-25 Tags: 
||      4. Björn Borg Phone: 98765432 Email: borg@gmail.com Address: 311, Clementi Ave 2, #02-25 Tags: 
|| 
|| 4 persons listed!
|| ===================================================
*****

***** actual.txt
|| Enter command: || [Command entered:  add José Eduardo p/98765432 e/borg@gmail.com a/311, Clementi Ave 2, #02-25]
|| Person names should be spaces or alphabetic characters
|| ===================================================
***** EXPECTED.TXT
|| Enter command: || [Command entered:  add José Eduardo p/98765432 e/borg@gmail.com a/311, Clementi Ave 2, #02-25]
|| New person added: José Eduardo Phone: 98765432 Email: borg@gmail.com Address: 311, Clementi Ave 2, #02-25 Tags: 
|| ===================================================
*****

***** actual.txt
||      3. John-Doe jr. Phone: 98765432 Email: johnd@gmail.com Address: 311, Clementi Ave 2, #02-25 Tags: 
|| 
[...]

PR #132, opened 2016-12-31
Change file extension requirement from .txt to .xml #130

zzzzwen:
Ready for review

need to manually delete the original `addressbook.txt` file from local disk to prevent git from uploading it since I have changed .gitignore

Checked actual running of add command, new `addressbook.xml` file generated.
StorageFileTest is working, test data also changed to .xml

---
This change is [Reviewable](https://reviewable.io/reviews/se-edu/addressbook-level2/132)


zzzzwen:
`expected.txt` and `input.txt` are not changed since I remembered that those are for regression test which only serves in level 1.

I think can remove the `runtests.bat` `runtests.sh` `expected.txt` and `input.txt`  right?

damithc:
@zzzzwen if you look at the level-2 dev guide, you'll see that both type of tests are still used.  
Are the tests passing? You need to run them locally as level 2 does not use gradle.

zzzzwen:
***** actual.txt
|| Enter command: || [Command entered:  add []\[;] p/12345 e/valid@e.mail a/vali
d, address]
|| Names should consist of letters, spaces, numbers, ', - or .
|| ===================================================
***** EXPECTED.TXT
|| Enter command: || [Command entered:  add []\[;] p/12345 e/valid@e.mail a/vali
d, address]
|| Person names should be spaces or alphabetic characters
|| ===================================================
*****

this is one error reported, the other one is .txt to .xml which I have fixed

this one also occurs in the master branch, I'm not sure why

pyokagan:
@zzzzwen 
> this error also occurs in the master branch,

Likely #117 . I provided a (possible) fix there, but don't have time to fully investigate. Maybe you could take over?

zzzzwen:
@damithc My laptop JVM does not have the issues in #117, but for the error I uploaded above, I went to search the `Person names should be spaces or alphabetic characters` in the project package but I cannot find this line in the whole project for both level1 and 2. I'm wondering if the `Messages` has been changed but not updated in the `expected.txt` file

zzzzwen:
@damithc  the commit which changes the message is #118 done by @Neurrone , should I follow this commit to change the `expected.txt` to the correct version?

damithc:
@zzzzwen It's better to report is a bug and create a PR to fix it.

zzzzwen:
ok, except this, anymore changes needed for this pr?

zzzzwen:
After merging with #135 , all tests passing in my computer currently

PR #138, opened 2017-01-07
Names with unicode characters are treated as invalid #117

### LINKED ISSUE #117: Names with unicode characters are treated as invalid

Neurrone:
Fixes #117. 

After reverting, only alphabets and space is accepted. Do we still want to allow some punctuation that are still ascii characters?

### DIFF
--- a/src/seedu/addressbook/data/person/Address.java
+++ b/src/seedu/addressbook/data/person/Address.java
@@ -30,7 +30,7 @@ public class Address {
     }
 
     /**
-     * Returns true if a given string is a valid person address.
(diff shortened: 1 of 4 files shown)


damithc:
`-` ,`,` and `/` are commonly used in English names, so I don't mind allowing it although I think users can live without them too. May be that can be a separate PR?

PR #214, opened 2017-09-06
[W4.3d][T09-A2] Lim Yong Song

limyongsong:
Created Classes for Block, Unit, PostalCode and Street that only hold values.
Address Class itself is not fundamentally changed.


src/seedu/addressbook/data/person/Address.java
```diff
@@ -22,13 +27,27 @@
      */
     public Address(String address, boolean isPrivate) throws IllegalValueException {
         String trimmedAddress = address.trim();
+        //Added for week4LO
+        String[] addressHolder = new String[99];
+        int i=0;
+       for(String retVal: trimmedAddress.split(",")){
```
limyongsong: Split loop is done with reference to https://www.tutorialspoint.com/java/java_string_split.htm

PR #238, opened 2017-09-11
[W5.4][W09-A3] Kelvin Lin

Esilocke:
Implemented optional data input to allow for incomplete data (e.g. valid phone and email, but no address), as addressbook currently requires valid data for all fields, and rejects them otherwise.


Esilocke:
Posted to wrong branch, please ignore this

PR #264, opened 2017-09-13
Confirmation

syy94:
Added a class ConfirmableCommand that if extended, prompts the user for confirmation. This will mainly be used for commands which have irreversible actions.


syy94:
pull request to wrong git

PR #302, opened 2018-01-20
Remove polymorphic behavior #37

### LINKED ISSUE #37: Remove polymorphic behavior
Polymorphic behavior should be removed as it is not one of the learning outcomes at level 2.
e.g.
execute()
toString()

eugenepeh:
**Work-in-progress** 

fix issue #37

### Development note: 
This issue seems to be very complicated. Initial plan to combine parsing and execution together in order to remove Command super class leads to problems such as unthorough ParserTest (as it is difficult to test parsing and execution separately), and also violation of SRP.

Alternate options:
1.  Make Command class absorb all the other Commands' behavior, and hence remove all the child Command classes such as AddCommand.
 E.g. under Command class, there will be all these methods such as performAdd, performDelete and etc, upon performing `execute()` determine the command type and perform the corresponding command action.
Con: may lead to many different constructors (belong to different type of commands), unused variables, a very long class and possibly violation of SRP.

2. Suggested by pyokagan - include all Command subclass into AggregateCommand as variable

Additional Notes:
Whichever option we choose, this will turn the code into a much uglier state than before.


pyokagan:
@eugenepeh Maybe something like this? That way you don't have to put everything in one class and break SRP.

```java
class AggregateCommand {
    AddCommand addCommand;
    ClearCommand clearCommand;
    ListCommand listCommand;
    ...

    AggregateCommand(AddCommand addCommand) {
        this.addCommand = addCommand;
    }

    AggregateCommand(ClearCommand clearCommand) {
        this.clearCommand = clearCommand;
    }

    public CommandResult execute() {
        if (addCommand != null) {
            return addCommand.execute();
        } else if (clearCommand != null) {
            return clearCommand.execute();
        } else ...
    }
}
```

eugenepeh:
> @eugenepeh Maybe something like this? That way you don't have to put everything in one class and break SRP.

Wow, that's brilliant! Never thought of doing it that way before, thanks @pyokagan!

eugenepeh:
Remark: not sure if I should rename Command to AggregateCommand in this PR because that might spoil some of the documentation.

eugenepeh:
Quite a bit PR... anyone would like to review? @yamgent @Zhiyuan-Amos ? @_@

src/seedu/addressbook/commands/Command.java
```diff
@@ -6,6 +6,8 @@
 
 import java.util.List;
 
+import javax.swing.text.View;
```
Zhiyuan-Amos: unused import statement :P

src/seedu/addressbook/commands/ExitCommand.java
```diff
@@ -16,6 +16,6 @@ public CommandResult execute() {
     }
 
     public static boolean isExit(Command command) {
-        return command instanceof ExitCommand; // instanceof returns false if it is null
+        return command.getCommandClass() == ExitCommand.class; // instanceof returns false if it is null
```
Zhiyuan-Amos: no more `instanceof` :P

eugenepeh:
> [3/9 & 6/9] are these commits required, since we are allowing for inheritance but not polymorphism i.e. we can still have protected methods and variables which can be called by subclasses?
Edit: It's used as part of learning objectives :P

Oh... okay sounds good, I will make the change and submit next iteration.

eugenepeh:
Ahh... upon reworking on it... I finally remembered why I removed the inheritance.

Suppose if inheritance is available - 
I initialized an `AddCommand` and pass into a method that requires the parameter `Command`. All the subclasses variables in `Command` will be uninitialized (null), and the `execute()` method wouldn't work properly.

Hence, I removed inheritance as a whole.

yamgent:
> I initialized an AddCommand and pass into a method that requires the parameter Command. All the subclasses variables in Command will be uninitialized (null), and the execute() method wouldn't work properly.
> 
> Hence, I removed inheritance as a whole.

Maybe that is because of the case whereby we mixed two "responsibilities" together?

1. `Command` as a way to skirt around overriding methods (avoiding Polymorphism)
1. `Command` as a foundation for sharing similar methods (Inheritance)

We can still make them inherit from another class (`BaseCommand`?), while still having the same `Command` class that we have in your current PR.

eugenepeh:
<pre>
Maybe that is because of the case whereby we mixed two "responsibilities" together?

Command as a way to skirt around overriding methods (avoiding Polymorphism)
Command as a foundation for sharing similar methods (Inheritance)
We can still make them inherit from another class (BaseCommand?), while still having the same Command class that we have in your current PR to get rid of polymorphism.
</pre>

Hmm, okay, I think I roughly get what you mean. Will try out that direction.

yamgent:
I did a quick UML diagram in case there is some misunderstanding. :P

remove polymorphism

src/seedu/addressbook/commands/AggregateCommand.java
```diff
@@ -0,0 +1,106 @@
...
+        } else if (viewAllCommand != null) {
+            return viewAllCommand.execute();
+        } else if (viewCommand != null) {
+            return viewCommand.execute();
+        } else {
+            throw new UnsupportedOperationException("None of the commands are valid.");
+        }
+    };
+
+    public Command getCommand() {
```
pyokagan: I think this is still taking advantage of polymorphism because we are referring to a subtype object via its supertype.

eugenepeh: Hmm, true. Guess I'll have to make it looks even uglier :P

pyokagan: Wait, nevermind, I take that back :-P The issue only talks about polymorphic behavior anyway, so I guess if you don't call any polymorphic methods everything is fine.

src/seedu/addressbook/commands/AggregateCommand.java
```diff
@@ -0,0 +1,106 @@
...
+    private DeleteCommand deleteCommand;
+    private ExitCommand exitCommand;
+    private FindCommand findCommand;
+    private HelpCommand helpCommand;
+    private IncorrectCommand incorrectCommand;
+    private ListCommand listCommand;
+    private ViewAllCommand viewAllCommand;
+    private ViewCommand viewCommand;
+
+    public AggregateCommand(AddCommand addCommand) {
```
pyokagan: Uh oh, I just realized this is actually ad-hoc polymorphism :-P

pyokagan: Method overloading.

pyokagan:
Something like the following to remove the method overloading? 

```java
class AggregateCommand {
    ...

    private AggregateCommand() {}

    public static AggregateCommand fromAddCommand(AddCommand addCommand) {
        AggregateCommand command = new AggregateCommand()
        command.addCommand = addCommand;
        return command;
    }

    ...
}
```

src/seedu/addressbook/Main.java
```diff
@@ -80,15 +81,15 @@ private void exit() {
...
     /** Reads the user command and executes it, until the user issues the exit command.  */
     private void runCommandLoopUntilExitCommand() {
-        Command command;
+        AggregateCommand aggregateCommand;
         do {
             String userCommandText = ui.getUserCommand();
-            command = new Parser().parseCommand(userCommandText);
-            CommandResult result = executeCommand(command);
+            aggregateCommand = new Parser().parseCommand(userCommandText);
+            CommandResult result = executeCommand(aggregateCommand);
```
damithc: Can we do this without using polymorphic variables too? I suspect we'll end up with lot of duplicate/boilerplate switch statements but that's the point we are trying to make. e.g.,
```java
switch (Parser.getCommandType(userCommandText)):
ADD:
    AddCommand addCommand = Parser.parseAddCommand(userCommandText);
    ...
DELETE:
   ...
``` 

Perhaps do a dirty PR (i.e., no need for commit organization etc.) to see how the result look like before deciding whether we really want to go that way?

On a related note, the need to make small steps through OOP concepts is less now, due to CS2030. We can decide to use all OOP concepts in level 2, including polymorphism. What do you guys think?

yamgent: > Can we do this without using polymorphic variables too? I suspect we'll end up with lot of duplicate/boilerplate switch statements but that's the point we are trying to make. e.g.,

Does that imply combining the parsing and the execution together? I think the reason we have `AggregateCommand` in the first place was because the commands aren't executed straight away, but is stored in a variable first, hence `AggregateCommand` was created to solve this "storing" issue.

> On a related note, the need to make small steps through OOP concepts is less now, due to CS2030. We can decide to use all OOP concepts in level 2, including polymorphism. What do you guys think?

You are right, I just checked the [web page for cs2030](https://nus-cs2030.github.io/1718-s2/schedule/index.html), they teach that in like week 2 :P.

I think it is OK to expect them to know polymorphism well given the syllabus in cs2030. So I think having a small codebase is more important here, in not making it too overwhelming for students.

Zhiyuan-Amos: I agree with Wang Leng, the lecture notes seem to be quite detailed too. The following image is from the CS2030 notes, which explains the need for polymorphism quite well already:

image

So I think there's perhaps no need for us to make our code to look something like the image above too :P

damithc: Sure we can leave it be. CS2030 is now a prerequisite for CS2103 and the number of students who took the old route is going to dwindle out in the coming semesters. @eugenepeh sorry I didn't think of this earlier. :-)

eugenepeh: @damithc oh okay, no problem. So this issue is closed?

damithc: Yes, please close both issue and PR. May want to point to this discussion when closing the issue.

PR #312, opened 2018-02-07
[W4.4][W14-A3]Randy Pang

randypx:
created Block, Street, Unit, PostalCode classes and update test cases


damithc:
This PR/issue should be against a repo in the org [nus-cs2103-AY1718S2](https://github.com/nus-cs2103-AY1718S2) not against the repo in se-edu org :-)
Closing...

PR #362, opened 2018-02-15
W5.11]t16 a1]ngo chanh anh minh

ncaminh:
Description:
.Small enhancement: I made the find command become non-case-sensitive.
.Big enhancement: I force the user to login to the address book with the correct ID and password. I save the ID and password in "adminInfo.txt"


khiayi:
Looks good. You have also added in the task from W4.4b, it should be written in your description as well.  👍

yamgent:
Hi guys, cs2103 PRs need to be opened against repos in [`nus-cs2103-AY1718S2`](https://github.com/nus-cs2103-AY1718S2/addressbook-level2), not here in the `se-edu` repo. :P

Closing...

PR #382, opened 2018-07-30
Ensure java 9 compatibility #381

### LINKED ISSUE #381: Ensure java 9 compatibility
Ensure that this project works with Java 9. Even better if we can have both Java 9 and 10 compatibility.

yamidark:
Fixes #381 
```
The project currently runs on Java 8.

Since AddressBookL4 is already running on Java 9, this project should
be upgraded to run on Java 9 as well for consistency.

Since the project has been adjusted to be Java 9 compatible in #388,
let's update the DeveloperGuide to mention using Java 9 or later.

Travis currently builds the project using JDK 8. Since we are now
running on Java 9, let's update it to build the project using JDK 9
as well.

UtilsTest uses the Integer(int) constructor for its equality test,
which is deprecated from Java 9[1]. Since use of deprecated APIs is
considered bad practice, let's replace this with Integer.valueOf(int).

[1] https://docs.oracle.com/javase/9/docs/api/java/lang/Integer.html#Integer-int-

### DIFF
--- a/docs/DeveloperGuide.adoc
+++ b/docs/DeveloperGuide.adoc
@@ -7,7 +7,7 @@
 
 *Prerequisites*
 
-* JDK 8 or later
+* JDK 9 or later
 * IntelliJ IDE
 
 *Importing the project into IntelliJ*
@@ -16,7 +16,7 @@
 . Open IntelliJ (if you are not in the welcome screen, click `File` > `Close Project` to close the existing project dialog first)
 . Set up the correct JDK version
 .. Click `Configure` > `Project Defaults` > `Project Structure`
-.. If JDK 8 is listed in the drop down, select it. If it is not, click `New...` and select the directory where you installed JDK 8.
+.. If JDK 9 is listed in the drop down, select it. If it is not, click `New...` and select the directory where you installed JDK 9.
 .. Click `OK`.
 . Click `Import Project`
 . Locate the project directory and click `OK`
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 matrix:
   include:
-    - jdk: oraclejdk8
+    - jdk: oraclejdk9
 
 script:
   - (cd test && ./runtests.sh) # I/O tests
@@ -18,4 +18,4 @@ deploy:
 addons:
   apt:
     packages:
-      - oracle-java8-installer
+      - oracle-java9-installer
--- a/test/java/seedu/addressbook/common/UtilsTest.java
+++ b/test/java/seedu/addressbook/common/UtilsTest.java
@@ -30,8 +30,8 @@ public class UtilsTest {
         assertNotUnique("abc", "abc");
         assertNotUnique("abc", "", "abc", "ABC");
         assertNotUnique("", "abc", "a", "abc");
-        assertNotUnique(1, new Integer(1));
-        assertNotUnique(null, 1, new Integer(1));
+        assertNotUnique(1, Integer.valueOf(1));
+        assertNotUnique(null, 1, Integer.valueOf(1));
         assertNotUnique(null, null);
         assertNotUnique(null, "a", "b", null);
     }

### CODE REVIEW

File: docs/DeveloperGuide.adoc [hunk 1]
```diff
@@ -16,24 +16,31 @@
...
 . Rename the project if you want. Click `Next`
 . Ensure that your `\src` and `\test\java` folders are checked. Keep clicking `Next`
 . Click `Finish`
+. Try running the application (right-click the `Main` class under the `src` folder, and click `Run 'Main.main()'`)
+. Observe how the application fails to compile. This is because the `JAXB` APIs is no longer contained under the default classpath from Java 9 onwards. To fix this issue, the `JAXB` module needs to be added to the classpath:
+.. Open any of the files which has the compilation error (e.g `StorageFile.java`), and place your cursor over any `javax.xml.bind` in the imports highlighted in red
+.. Press ALT+ENTER and select `--add-modules=java.xml.bind to module compiler options`
+.. Go to `Run` -> `Edit Configurations...`
+.. Under `Configuration`, add `--add-modules java.xml.bind` into `VM options`
+. Run the application again to ensure it compiles and runs without any errors.
```
yamgent: Remove the full stop at the end of this sentence, to be consistent with the rest of the list items.

docs/DeveloperGuide.adoc
```diff
@@ -16,24 +16,31 @@
...
+.. If JDK 9 is listed in the drop down, select it. If it is not, click `New...` and select the directory where you installed JDK 9.
 .. Click `OK`.
 . Click `Import Project`
 . Locate the project directory and click `OK`
 . Select `Create project from existing sources` and click `Next`
 . Rename the project if you want. Click `Next`
 . Ensure that your `\src` and `\test\java` folders are checked. Keep clicking `Next`
 . Click `Finish`
+. Try running the application (right-click the `Main` class under the `src` folder, and click `Run 'Main.main()'`)
+. Observe how the application fails to compile. This is because the `JAXB` APIs is no longer contained under the default classpath from Java 9 onwards. To fix this issue, the `JAXB` module needs to be added to the classpath:
```
yamgent: I played around with the codebase a little more.

Actually, we can fix the issue of the missing modules on our side, by adding a `module-info.java` to `src/`:

```java
open module addressbook.level2 {
    requires java.xml.bind;
    exports seedu.addressbook.storage.jaxb;
}
```

After doing that, students need not add `--add-modules` manually as specified in the instructions anymore.

This doesn't seem to work on tests though, not sure if `module-info.java` can be used on tests as well? Or perhaps it is placed wrongly?

Also not sure whether this entire solution is the best to begin with. What is your opinion on this?

yamidark: Based on this [issue](https://youtrack.jetbrains.com/issue/IDEA-171419) with Intellij, seems like the tests are run on the classpath instead of the modulepath, so it is unable to get the modules in `module-info.java`.

docs/DeveloperGuide.adoc
```diff
@@ -28,12 +28,15 @@
 .. Open any test file in `\test\java` and place your cursor over any `@Test` highlighted in red
 .. Press ALT+ENTER and select `Add 'JUnit4' to classpath`
 .. Select `Use 'JUnit4' from IntelliJ IDEA distribution` and click `OK`
+. Add dependency on module `addressbook-level2`
+.. Open any test file in `\test\java` and place your cursor over any imports from `seedu.addressbook` highlighted in red
+.. Press ALT+ENTER and select `Add dependency on module 'addressbook-level2'`
```
yamgent: I wasn't able to get this option (`seedu.addressbook` doesn't highlight in red for me, yet execution of the test still fails anyway due to the missing module).

docs/DeveloperGuide.adoc
yamgent: > Although, there will still be a need to specify the --add-modules java.xml.bind flag (or the module path) in the command line and for the tests.

Yeah, I don't think there's a way around that.

docs/DeveloperGuide.adoc
yamgent: You might want to try deleting the `.idea` folder, or re-clone the project, and try the instructions again. The existing configuration might be interfering with the how you import the modules.

yamgent:
As discussed with Prof @damithc, `java.xml.bind` is [marked as deprecated since Java 9](https://docs.oracle.com/javase/9/docs/api/java.xml.bind-summary.html), and according to an answer on stackoverflow, [they are planning to remove it in Java 11](https://stackoverflow.com/a/43574427). Using deprecated modules is a bad practice, and we certainty don't want the students to adopt that.

An alternative way discussed, was to find out whether we can use third party libraries for our XML needs. We have to do this without using gradle (so we will have to tell the students how they should install the third party library if necessary).

@pyokagan do you have any input regarding this issue?

pyokagan:
> We have to do this without using gradle (so we will have to tell the students how they should install the third party library if necessary).

Yeah, and vendoring in dependencies is Bad Practice(tm).

Maybe reuse the storage code from AB-1? JAXB and its uses of introspection and annotations seems to be too advanced for a "level 2" anyway.

I do think sooner or later students should start using Gradle from AB-3 though -- JavaFX itself will soon only be available as a dependency and not distributed with the JDK.

yamidark:
>Maybe reuse the storage code from AB-1? JAXB and its uses of introspection and annotations seems to be too advanced for a "level 2" anyway.

Hmm, should this be done within this PR though, since it is part of ensuring the Java 9 compatibility, or in a different PR, as it may be considered to be a different issue?

yamgent:
> Hmm, should this be done within this PR though, since it is part of ensuring the Java 9 compatibility, or in a different PR, as it may be considered to be a different issue?

Some of these changes won't be relevant with the storage system change though. Prof mentioned that we should consider the alternate storage system unless there isn't time to finish the implementation on time.

You can close this PR and do the storage system change in another PR.

yamidark:
>You can close this PR and do the storage system change in another PR.

Can put this PR on hold 1st, will do the storage system change 1st before updating the documentations for Java 9 here.

damithc:
> Can put this PR on hold 1st, will do the storage system change 1st before updating the documentations for Java 9 here.

It's good if we can do this by week 1. This batch of students might go through AB1-4 faster than previous batches.

yamidark:
Continuing on this since #388 has been merged.

yamgent:
Instructions and codebase looks fine. :+1:

I am not sure whether it is because I updated my IDE recently, or whether it is because we migrated to Java 9, but this step in the setup instruction no longer match for me:

https://github.com/se-edu/addressbook-level2/blob/3a88b067b25d5845316f157251798903965637ad/docs/DeveloperGuide.adoc#L30

I actually get this dialog box instead.



Do you get this dialog box too?

yamidark:
>Do you get this dialog box too?

Yep, got this too. Probably due to a newer Intellij version since setting up using Java 8 gives the same dialog box. Will update the instructions.

PR #385, opened 2018-08-06
AddressBook: remove master tag list #383

### LINKED ISSUE #383: AddressBook: remove master tag list
We keep track of a master tag list which holds all the tags in
AddressBook.

As the master tag list was deemed unnecessary and may be too 
complex for new developers to grasp, as discussed in
se-edu/addressbook-level4#753 and se-edu/addressbook-level4#794,
the master tag list has already been removed in AB4
by se-edu/addressbook-level4#824. As such, the master tag list
should also be removed for the lower level projects.

Let's remove the master tag list in AddressBook.

yamidark:
Fixes #383
```
We keep track of a master tag list which holds all the tags in
AddressBook.

As the master tag list was deemed unnecessary and may be too
complex for new developers to grasp, as discussed in
se-edu/addressbook-level4#753 and se-edu/addressbook-level4#794,
the master tag list has already been removed in AB4
by se-edu/addressbook-level4#825. 

UniqueTagList was also deemed unnecessary and removed in
addressbook-level4 after the master tag list is removed.

As such, the master tag list and UniqueTagList class should also be
removed in the lower level projects.

Let's remove the master tag list in AddressBook and AdaptedAddressBook,
as well as the UniqueTagList class.
```


test/java/seedu/addressbook/commands/AddCommandTest.java
```diff
@@ -26,7 +26,7 @@
 
 public class AddCommandTest {
     private static final List<ReadOnlyPerson> EMPTY_PERSON_LIST = Collections.emptyList();
-    private static final Set<String> EMPTY_STRING_LIST = Collections.emptySet();
+    private static final Set<String> EMPTY_STRING_LIST = new HashSet<>();
```
eugenepeh: is this change necessary? i think the intention was to be immutable.

yamidark: Hmm, I think I did a mass replacement for this, didn't notice it changed here as well.
Anyway, since the `Person` constructor now directly sets the `tags`, thought it would be safer to ensure that the `Set` we supply is mutable from the start.
Perhaps would be better if we make `Person` hold its own mutable copy `Set` from the start, and make it add the `Tag`s in, like in `UniqueTagList`?

pyokagan:
@yamidark Can squash [1/4] and [2/4]? [1/4] makes no sense without [2/4], as [2/4] deletes huge chunks of code that were touched in [1/4].

I'm not quite sure about the deletions of some test cases in `AddressBookTest`, like `addPerson_emptyAddressBook`, but I also don't quite mind given the time constraints.

btw, [1/4]'s commit message:

> ```
> AddressBook: remove master tag list
> 
> We keep track of a master tag list which holds all the tags in
> AddressBook.
> 
> As the master tag list was deemed unnecessary and may be too complex
> for new developers to grasp, as discussed in
> se-edu/addressbook-level4#753 and se-edu/addressbook-level4#794,
> the master tag list has already been removed in addressbook-level4
> in se-edu/addressbook-level4#824. As such, the master tag list
> ```

Should be se-edu/addressbook-level4#825 I think.

> ```
> should also be removed here for consistency.
> 
> Let's remove the master tag list in AddressBook.
> ```

src/seedu/addressbook/data/person/Person.java
```diff
@@ -15,16 +18,24 @@
...
      * Assumption: Every field must be present and not null.
      */
-    public Person(Name name, Phone phone, Email email, Address address, UniqueTagList tags) {
+    public Person(Name name, Phone phone, Email email, Address address, Set<Tag> tags) {
         this.name = name;
         this.phone = phone;
         this.email = email;
         this.address = address;
-        this.tags = new UniqueTagList(tags); // protect internal tags from changes in the arg list
+        this.tags = tags;
```
pyokagan: Should be:

```java
this.tags = new HashSet<>(tags);
```

so that the `Person` will have a `Set` which it has full ownership of, and can't be modified externally (preserves encapsulation).

Can copy from AB-4:

https://github.com/se-edu/addressbook-level4/blob/e186102b646d2307b53928142dd22932a6f4c86d/src/main/java/seedu/address/model/person/Person.java#L25-L37

test/java/seedu/addressbook/commands/DeleteCommandTest.java
```diff
@@ -33,14 +33,14 @@
     @Before
     public void setUp() throws Exception {
         Person johnDoe = new Person(new Name("John Doe"), new Phone("61234567", false),
-                new Email("john@doe.com", false), new Address("395C Ben Road", false), new UniqueTagList());
+                new Email("john@doe.com", false), new Address("395C Ben Road", false), new HashSet<>());
```
pyokagan: Does the new constructor you introduced help here?

```java
public Person(Name name, Phone phone, Email email, Address address, Tag... tags)
```

test/java/seedu/addressbook/util/TypicalPersons.java
```diff
@@ -20,13 +20,13 @@
...
+                    new Address("1 Clementi Road", false), new HashSet<>());
             bill = new Person(new Name("Bill Clint"), new Phone("92229222", false), new Email("bc@gmail.com", false),
-                    new Address("2 Clementi Road", true), new UniqueTagList());
+                    new Address("2 Clementi Road", true), new HashSet<>());
             candy = new Person(new Name("Candy Destiny"), new Phone("93339333", true),
-                    new Email("cd@gmail.com", false), new Address("3 Clementi Road", true), new UniqueTagList());
+                    new Email("cd@gmail.com", false), new Address("3 Clementi Road", true), new HashSet<>());
             dan = new Person(new Name("Dan Smith"), new Phone("1234556", true), new Email("ss@tt.com", true),
-                    new Address("NUS", true), new UniqueTagList(new Tag("Test")));
+                    new Address("NUS", true), new HashSet<>());
```
pyokagan: Why was the `new Tag("Test")` removed?

pyokagan:
This wasn't very fun to review due to all the subtle differences between the different ABs. \</rant>

test/java/seedu/addressbook/util/TypicalPersons.java
yamidark: Whoops, didn't notice this

test/java/seedu/addressbook/commands/DeleteCommandTest.java
pyokagan: On second thought, I think you should remove the constructor. AB-4 doesn't have it.

yamidark: The new constructor was added to replace the `UniqueTagList` constructor:
```Java
public UniqueTagList(Tag... tags)
```
which AB4 doesn't have.
Although this is only used in tests, so I wouldn't really mind removing it as well to be consistent with AB4.

test/java/seedu/addressbook/data/AddressBookTest.java
```diff
@@ -48,25 +51,25 @@ public void setUp() throws Exception {
...
                                     new Address("10 Science Drive", false),
-                                    new UniqueTagList(tagScientist));
+                                    Collections.singleton(tagScientist));
 
         davidElliot    = new Person(new Name("David Elliot"),
                                     new Phone("84512575", false),
                                     new Email("douglas@nuscomputing.com", false),
                                     new Address("11 Arts Link", false),
-                                    new UniqueTagList(tagEconomist, tagPrizeWinner));
+                                    new HashSet<>(Arrays.asList(tagEconomist, tagPrizeWinner)));
```
eugenepeh: [2/4] If want to make it immutable, `Set.of(...)` would probably be a better option.

yamidark: This is currently still in Java 8, so `Set.of(...)` is still not available yet 😞.

eugenepeh: oh... crap lol

yamidark:
>eh... how come tag used a special purple color?

`Tag` is now strongly part of `Person`, so should have the same colour as the rest of the fields.
Although not too sure if should keep the original colour for the `LO` `Taggings` =x.

eugenepeh:
> Tag is now strongly part of Person, so should have the same colour as the rest of the fields.
Although not too sure if should keep the original colour for the LO Taggings =x (yep, looks weird with that single purple box inside).

I thought it was originally orange?!

yamidark:
>I thought it was originally orange?!

Yep, I changed it to purple to be consistent with the rest of the `Person` fields (in DeveloperGuide).

eugenepeh:
> Oh, just noticed that DeveloperGuide wasn't updated in this PR at all, just changed the diagram, since you probably didn't notice it.

Thanks, didnt notice that.

In this case... I think this line `Note how the setTags method of the Person class cannot be converted to a class-level method.` in LO is obsolete?

yamidark:
>Note how the setTags method of the Person class cannot be converted to a class-level method.

Ehh, as the method is modifying the instance-level variable `tags`, the `setTags` method still cannot be converted to class-level right?

eugenepeh:
> Ehh, as the method is modifying the instance-level variable tags, the setTags method still cannot be converted to class-level right?

PS, I actually mean that that line doesn't really serve a purpose anymore because prior to this change, it is distinct from the other members because we have a master list. But now it is no different from name... address... etc. So it brings up question like `whats so special about setTags`?

yamidark:
>Perhaps, we should rephrase it abit to like For example, instead of Note?

Yep, sounds good. Will do that in the next iteration.

pyokagan:
[3/4]

> ```
> Tag: move Tag from data/tag to data/person package
> 
> AddressBook used to have a master tag list, which stores
> the tag held by each Person.
> 
> As the master tag list has been removed, Tag is now strongly
> part of Person. As such, Tag should be under data/person
> package as well.
> ```

Maybe, but how does this weigh against keeping consistent with AB-4 (which does not have such a change?)

Let's refrain from making too many creative changes and drop [3/4]? We are pretty short of time.

PR #392, opened 2018-08-26
DeveloperGuide: add solution to src being treated as normal folder

eugenepeh:
Similar fix in https://github.com/se-edu/addressbook-level1/pull/105

```
Fix runtests scripts proceeding on compilation error
Occasionally, new developers may accidentally import the project as
a Gradle project.

This cause IntelliJ to often triggers Gradle synchronization which
allows Gradle to impose the project to use its configuration. Due
to this issue, developers will encounter the problem of marked
directory keep getting reset upon restarting IntelliJ.

Let's add the solution to this problem in the troubleshooting
section to help the developers who have accidentally misconfigured
the importation.
```

### DIFF
--- a/docs/DeveloperGuide.adoc
+++ b/docs/DeveloperGuide.adoc
@@ -64,7 +64,7 @@ the test has passed
 
 * In IntelliJ, right-click on the `test` folder and choose `Run 'All Tests'`
 
-=== Troubleshooting test failures
+=== Troubleshooting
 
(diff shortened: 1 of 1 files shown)


yamgent:
As stated in https://github.com/se-edu/addressbook-level1/pull/105#issuecomment-416036095 you might want to add the uncheck instruction back in.

eugenepeh:
Since AB2 has the gradle setup required, this should not happen. 
Closing this PR.

PR #397, opened 2018-09-04
[LO-Encapsulation][W13-3]Andrew Tan Jin Shen

andrewtanJS:
Encapsulate CommandResult


andrewtanJS:
I may have jumped the gun by doing the encapsulation LO under the JUnit test LO with the intent on simply completing a PR

PR #401, opened 2018-09-05
[W4.11e][T13-2]Daniel Khoo

DanKhoo:


damithc:
This PR/issue should be against a repo in the org [nus-cs2103-AY1819S1](https://github.com/nus-cs2103-AY1819S1) not against the repo in se-edu org :-)
Closing...

PR #407, opened 2018-09-05
UtilsTest.java: added tests for isAnyNull method

abtyx:
added Test cases which utilize the Utils#isAnyNull method


eugenepeh:
@abtyx note that your PR is supposed to be targeted against the `module organisation`'s repository and not under `se-edu`. 

Do kindly submit your work against the right upstream repository.
We will be closing this PR.

PR #418, opened 2018-09-07
Unit test

Hexinyiyi:
Submit Junit test


eugenepeh:
@Hexinyiyi note that your PR is supposed to be targeted against the `module organisation`'s repository and not under `se-edu`. 

Do kindly submit your work against the right upstream repository.
We will be closing this PR.

PR #427, opened 2018-09-08
Edit find function

clawyq:
Edit find function to not be case sensitive
Update user guide to reflect the change in the find method
Edit JUnit tests to reflect the changes in the behaviour of the new find method
Edit test files to reflect the changes in the behaviour of the new find method


damithc:
This PR/issue should be against a repo in the org [nus-cs2103-AY1819S1](https://github.com/nus-cs2103-AY1819S1) not against the repo in se-edu org :-)
Closing...

PR #431, opened 2018-09-09
ParserTest: Fix bug in String.format

sijie123:
**This is not a submission for an LO. It is a pull request for an actual bug fix.**

In the current implementation of ParserTest, parse_addCommandInvalidArgs_errorMessage and parse_addCommandInvalidPersonDataInArgs_errorMessge are implemented wrongly. Both used String.format with $s instead of %s. 

String.format actually does not format the string correctly with Name.EXAMPLE, Phone.EXAMPLE etc.
For example, 
[Line 237] String.format(addCommandFormatString, invalidName, validPhoneArg, validEmailArg) becomes "add $s $s $s a/123, some street". You can verify this by setting a breakpoint in the debugger. 
The expected behaviour should be "add []\[;] p/123456789 e/valid@e.mail a/123, some street.

All the tests are currently passing because these two methods are supposed to test for InvalidCommand, which returns the correct result despite a faulty test input.
On the other hand, parse_addCommandValidPersonData_parsedCorrectly works as intended as well because it generatesTestPerson then converts by concatenating strings, which subverts the bug too.

### DIFF
--- a/test/java/seedu/addressbook/parser/ParserTest.java
+++ b/test/java/seedu/addressbook/parser/ParserTest.java
@@ -208,11 +208,11 @@ public class ParserTest {
             "add ",
             "add wrong args format",
             // no phone prefix
-            String.format("add $s $s e/$s a/$s", Name.EXAMPLE, Phone.EXAMPLE, Email.EXAMPLE, Address.EXAMPLE),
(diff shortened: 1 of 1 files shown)


sijie123:
Closing this PR - Accidentally deleted the previous forked repo and can't push to the PR to make changes... (Refer to #472)

PR #433, opened 2018-09-10
Parser: replace regular expressions with a simpler method #425

### LINKED ISSUE #425: Parser: replace regular expressions with a simpler method
Current: regular expressions are used to do the basic parsing of the command

Suggested: use a simpler method

Justification: AB-2 is targetted at students who are just beginning to learn OOP concepts. Regular expressions technique does not match the level of sophistication we can expect from an average student at that level.

apoorva17:
Fixes #425 

Regex is being used to parse commands currently. 

In order to simplify the understanding of Parser for students, let's replace regex with simpler methods in Parser.

### DIFF
--- a/src/seedu/addressbook/parser/Parser.java
+++ b/src/seedu/addressbook/parser/Parser.java
@@ -29,9 +29,7 @@ import seedu.addressbook.data.exception.IllegalValueException;
  * Parses user input.
  */
 public class Parser {
-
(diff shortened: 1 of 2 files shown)


apoorva17:
@devs ready for review!

damithc:
@apoorva17 see if you can replicate the AB-4 solution here.

PR #436, opened 2018-09-10
Utils.isAnyNull: Add checks on the entire object to cater for null parameter

### LINKED ISSUE #426: Utils.isAnyNull: cater for null parameter
Current: `Utils.isAnyNull(null)` throws an NPE (but `Utils.isAnyNull((Object)null);` works)
Suggested: return `true`

sijie123:
Fixes #426 

Add an extra check for null.
Returns true if parameter passed in is null.


damithc:
@sijie123 you can edit commits and force-push to update the PR. No need to close the PR and create a new one.

PR #459, opened 2018-09-12
add count command

hotarumin:


CrimsonAng:
It's nice to see that you created a new variable/method dedicated for counting itself so that people who read the code would understand the purpose of the variable. 

Some people might be confused if the coder just extrapolated number of people from the 'list' command instead of having a segregated variable/method.

damithc:
This PR/issue should be against a repo in the org of your module e.g.,  [nus-cs2103-AY1819S1](https://github.com/nus-cs2103-AY1819S1) or [nusCS2113-AY1819S1](https://github.com/nusCS2113-AY1819S1/), not against the repo in se-edu org :-)
Closing...

PR #469, opened 2018-09-21
Developer Guide: Fix spelling error

### LINKED ISSUE #289: Fix spelling error in DeveloperGuide#design
It has been pointed out by the CS2103 students that the spelling of `ClearCommand` in the [design](https://github.com/se-edu/addressbook-level2/blob/master/doc/DeveloperGuide.md#design) is wrong

apoorva17:
Fixes #289 
```
Spelling of ClearCommand was incorrect in Design.

Let's correct the spelling.
```


pyokagan:
@apoorva17 Thanks for the PR.

No need to send a new iteration for this since the commit message is can still be understood, but just a note on the commit message for future PRs:

> ```
> Developer Guide: Fix spelling error
> 
> Spelling of ClearCommand was incorrect in Design.
> ```

The commit message is set in the point in time *before* the commit is applied.

In other words, we should use present tense when describing the state of the code base *before* the commit. So, it should be: "Spelling of ClearCommand *is* incorrect in Design."
> ```
> Let's correct the spelling.
> ```

This is why we also use imperative mood when describing what the commit will do when applied, as you have done here.

PR #485, opened 2019-02-10
Fix file access errors in runtests.bat when case sensitive attribute is enabled on directories

Creastery:
It appears that sub-problem 2 of se-edu/addressbook-level1#111 (`fc` reporting a "no such file or folder" error) is applicable to AB2 as well.

May I suggest the [same fix](https://github.com/se-edu/addressbook-level1/pull/112#issuecomment-457542443) of renaming `actual.txt` and `expected.txt` to uppercase, like in se-edu/addressbook-level1@016474ea3dcd2fb45a1c7c7d9f936f0371e04e64?


damithc:
Thanks for the fix @Creastery

Issue #503, opened 2019-05-16
JDK version is no longer supported.


j-lum:
AB-2 is still using JDK8 and support for JDK9 from Oracle is no longer available. We should upgrade to the next available LTS version.

PR #504, opened 2019-05-16
[#503]Migration to JDK11+

j-lum:
For consistency with the other ABs, let's update AB-2 to JDK11.
Fortunately the codebase can be ported without any changes.

[1/2] Updated travis to use Oracle JDK 11
[2/2] Updated references to JDK 8 to JDK 11 in the Developer Guide.

### DIFF
--- a/docs/DeveloperGuide.adoc
+++ b/docs/DeveloperGuide.adoc
@@ -8,7 +8,7 @@
 
 *Prerequisites*
 
-* JDK 8
+* JDK 11
 * IntelliJ IDE
 
 *Importing the project into IntelliJ*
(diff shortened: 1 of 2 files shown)


pyokagan:
@j-lum 
> [1/2] Updated travis to use Oracle JDK 11
> [2/2] Updated references to JDK 8 to JDK 11 in the Developer Guide.

You don't need to add this to the PR description ;-) In fact, most of the time I don't even look at the PR description since it's not entered into Git history.

pyokagan:
> In fact, most of the time I don't even look at the PR description since it's not entered into Git history.

Still, I must still quantify this by saying that PR descriptions can still be useful for PRs with multiple commits ;-)

docs/DeveloperGuide.adoc
```diff
@@ -8,16 +8,16 @@
 
 *Prerequisites*
 
-* JDK 8
-* IntelliJ IDE
+* JDK 11
+* IntelliJ IDEA
```
pyokagan: AB-1, AB-3 and AB-4 all use "IntelliJ IDE", so I'm not sure if it makes sense to make AB-2 inconsistent with the others, especially since this PR is just supposed to do maintenance. Shall we remove this change?

j-lum: Let's move postpone this change to the documentation overhaul later on. Strictly speaking `IntelliJ IDEA` is the name of the IDE but that's a discussion that we can have in a separate PR.

pyokagan: Yes, it is important to strictly separate commits between those that are simply performing maintenance (refactoring etc.) and enhancements.

PR #506, opened 2019-05-23
Migrate to JUnit 5 from JUnit 4

j-lum:
Change imports to JUnit 5 imports.

Replace `@Before` with `@BeforeEach`, `ExpectedException` with `assertThrows`,
`TemporaryFolder` with`TempDir` where applicable as the former are
depreciated in JUnit 5.

Add dependencies for JUnit 5 to build.gradle.txt.


test/java/seedu/addressbook/commands/DeleteCommandTest.java
```diff
@@ -21,6 +22,7 @@
 import seedu.addressbook.ui.TextUi;
 import seedu.addressbook.util.TestUtil;
 
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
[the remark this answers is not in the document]
j-lum: Not necessary since leaving it out will default back to JUnit 4 behavior. 

However, there are some nice benefits to having it in as described [here](https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-instance-lifecycle) and it has the nice side effect of making the tests go quite a bit faster.

docs/DeveloperGuide.adoc
```diff
@@ -25,10 +25,10 @@
 . Rename the project if you want. Click `Next`
 . Ensure that your `\src` and `\test\java` folders are checked. Keep clicking `Next`
 . Click `Finish`
-. Add JUnit 4 to classpath
+. Add JUnit 5 to classpath
 .. Open any test file in `\test\java` and place your cursor over any `@Test` highlighted in red
 .. Press kbd:[ALT + ENTER] and select `Add 'JUnit4' to classpath`
```
pyokagan: There's still a reference to JUnit4 here. Intended?

pyokagan: On my side, it says "Add 'JUnit5.3' to classpath"

docs/DeveloperGuide.adoc
```diff
@@ -25,10 +25,10 @@
...
-.. Select `junit:junit:4.12` and click `OK` (if `4.12` is not available, choose any `4.x` that is later than `4.12`)
+.. Select `junit:junit:5.40` and click `OK` (if `5.40` is not available, choose any `5.4x` that is later than `5.40`)
```
pyokagan: `junit:junit:5.40` should not be possible since JUnit5 is using a different artifact/group name on maven.

Maybe run through the instructions in IntelliJ? It says `org.junit.jupiter:junit-jupiter-api:RELEASE` on my side.

docs/DeveloperGuide.adoc
j-lum: On IDEA 2018.2, it shows as JUnit 5.2 so I'm pretty sure it depends on the version of IDEA you have. 
Should we just change it to be more generic?

docs/DeveloperGuide.adoc
j-lum: I was actually thinking that they should hit the refresh button and select 5.4.x from the dropdown since RELEASE might point to a release that breaks later on.

docs/DeveloperGuide.adoc
pyokagan: Yes we can probably follow the "(if `5.2` is not available, choose any `5.x` that is later than `5.2`)" wording used below.

docs/DeveloperGuide.adoc
pyokagan: > I was actually thinking that they should hit the refresh button and select 5.4.x from the dropdown since RELEASE might point to a release that breaks later on.

Oh Okay I can see the list of versions now after waiting (a long time for some reason).

j-lum: Yeah on different versions of of IDEA the list fetch varies in size as well. Kind of annoying, should we consider putting a screenshot in? Messing this step up could result in a fair bit of trouble for students.

pyokagan: > Kind of annoying, should we consider putting a screenshot in?

ok.

Another idea: how about just asking students to replace `org.junit.jupiter:junit-jupiter-api:RELEASE` with `org.junit.jupiter:junit-jupiter-api:5.4.0` and then pressing OK directly? That way we won't need to depend on IntelliJ getting the list correct, and it's just a copy and paste.

docs/DeveloperGuide.adoc
j-lum: Wait actually that's the section header to the part where we instruct them how to get the correct version of JUnit so that should actually be `Add JUnit5.4 to classpath` right?

pyokagan: No, I don't think it's the section header.

pyokagan:
I'll request @damithc to review as well.

docs/DeveloperGuide.adoc
```diff
@@ -25,10 +25,14 @@
...
 . Ensure that your `\src` and `\test\java` folders are checked. Keep clicking `Next`
 . Click `Finish`
-. Add JUnit 4 to classpath
+. Add JUnit 5 to classpath
 .. Open any test file in `\test\java` and place your cursor over any `@Test` highlighted in red
-.. Press kbd:[ALT + ENTER] and select `Add 'JUnit4' to classpath`
-.. Select `junit:junit:4.12` and click `OK` (if `4.12` is not available, choose any `4.x` that is later than `4.12`)
+.. Press kbd:[ALT + ENTER] and select `Add 'JUnit5.3' to classpath` (if `5.3` is not available, any `5.x` will suffice)
++
+image::HelpfulIDEA.png[]
```
pyokagan: Also, I wonder if a filename that describes the relevant steps in the documentation would be more helpful, e.g. (`AltEnterSelectJunit53.png`)

j-lum:
Snuck in a quick rename of the image.

pyokagan:
@j-lum Any comments on https://github.com/se-edu/addressbook-level2/pull/506#pullrequestreview-242345052 ?

j-lum:
> 
> 
> @j-lum Any comments on [#506 (review)](https://github.com/se-edu/addressbook-level2/pull/506#pullrequestreview-242345052) ?

Might be better to not implicitly endorse a JUnit release candidate since the purpose of the screenshot is to demonstrate the usage of the dropdown list.

PR #507, opened 2019-05-24
Migrate to checkstyle 8.2

j-lum:
Upgrade checkstyle version and checkstyle rules to match that of later
AddressBooks for consistency.

Add checkstyle checks into the buildfile to ensure that future commits
will remain compliant to checkstyle rules.

Fix checkstyle violations as the AddressBooks are meant to be models for
students to learn from.

Modify checkstyle rules to allow JUnit4 annotations until the migration
to JUnit5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
 
 script:
   - (cd test && ./runtests.sh) # I/O tests
-  - ./gradlew -b build.gradle.txt clean test # JUnit tests
+  - ./gradlew -b build.gradle.txt clean test checkstyleMain checkstyleTest # JUnit test and runs checkstyle
   - ./gradlew -b build.gradle.txt asciidoctor # Documentation
 
 deploy:
--- a/src/seedu/addressbook/commands/CommandResult.java
+++ b/src/seedu/addressbook/commands/CommandResult.java
@@ -1,9 +1,9 @@
 package seedu.addressbook.commands;
 
-import seedu.addressbook.data.person.ReadOnlyPerson;
-
 import java.util.List;
 import java.util.Optional;
+
+import seedu.addressbook.data.person.ReadOnlyPerson;
 
 /**
  * Represents the result of a command execution.
--- a/src/seedu/addressbook/ui/TextUi.java
+++ b/src/seedu/addressbook/ui/TextUi.java
@@ -21,6 +21,9 @@ import seedu.addressbook.data.person.ReadOnlyPerson;
  */
 public class TextUi {
 
+    /** Offset required to convert between 1-indexing and 0-indexing.  */
+    public static final int DISPLAYED_INDEX_OFFSET = 1;
+
     /** A decorative prefix added to the beginning of lines printed by AddressBook */
     private static final String LINE_PREFIX = "|| ";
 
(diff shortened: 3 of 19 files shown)


src/seedu/addressbook/ui/TextUi.java
```diff
@@ -91,7 +90,12 @@ public String getUserCommand() {
         return fullInputLine;
     }
 
-
+    /**
+     * Helper method to generate and print the welcome message upon the start of
```
damithc: The convention is to start with a verb. e.g., `Generates and prints ...`

test/java/seedu/addressbook/storage/StorageFileTest.java
```diff
@@ -21,8 +23,7 @@
 import seedu.addressbook.data.person.Phone;
 import seedu.addressbook.data.tag.Tag;
 import seedu.addressbook.storage.StorageFile.StorageOperationException;
-import static seedu.addressbook.util.TestUtil.assertTextFilesEqual;
-import static seedu.addressbook.util.TestUtil.assertFileDoesNotExist;
+
```
pyokagan: This extra line seems unnecessary?

test/java/seedu/addressbook/util/TypicalPersons.java
```diff
@@ -34,6 +37,11 @@ public TypicalPersons() {
         }
     }
 
+    /**
+     * Inserts predefined {@code Person} objects in to a given instance of
```
pyokagan: Should be `into`?

test/java/seedu/addressbook/util/TypicalPersons.java
```diff
@@ -34,6 +37,11 @@ public TypicalPersons() {
...
+     * {@code AddressBook}
```
pyokagan: Can be in one line?

src/seedu/addressbook/commands/CommandResult.java
```diff
@@ -1,10 +1,10 @@
 package seedu.addressbook.commands;
 
-import seedu.addressbook.data.person.ReadOnlyPerson;
-
 import java.util.List;
 import java.util.Optional;
 
+import seedu.addressbook.data.person.ReadOnlyPerson;
+
```
pyokagan: Seems like an unnecessary new line?

test/java/seedu/addressbook/util/TypicalPersons.java
pyokagan: Side thought that doesn't need to be acted on at all: Isn't it more important that we document the *public* methods `getTypicalPersons()` and `getTypicalAddressBook()` rather than documenting this? :thinking:

j-lum: I would honestly rather not be writing JavaDocs for simple, short functions...

j-lum:
Oops, added the check task to the travis build.

.travis.yml
```diff
@@ -8,7 +8,7 @@ install:
 
 script:
   - (cd test && ./runtests.sh) # I/O tests
-  - ./gradlew -b build.gradle.txt clean test # JUnit tests
+  - ./gradlew -b build.gradle.txt clean test check # JUnit tests
```
pyokagan: `check` includes `test` as well, so `test` is redundant if we are running `check`.

Also, is the comment "JUnit tests" still correct if we are running checkstyle as well?

j-lum: Actually let's change `check` to `checkstyleMain` and `checkstyleTest`. `check` itself is an aggregate task that depends on `test` so I thought it would be better to be more explicit. Better to leave no room for confusion.

pyokagan: > check itself is an aggregate task that depends on test so I thought it would be better to be more explicit. Better to leave no room for confusion.

Actually, it will confuse students more since they would assume that `check` will not run `test`.

Also, as a general rule we always write code that is succinct, because: (1) It demonstrates that you understand what is going on, and (2) so we write code which do the same thing the same way (code style standardization).

